PR #19, opened 2017-11-30
Make converters module a package

delta2323:
This PR changes `chainerchem.dataset.conveters` from a module (`converters.py`) to a package (`_converters` directory), moves `concat_mols` function under the namespace.


delta2323:
Closed, as we will consider it again when we add new converters.

PR #27, opened 2017-12-04
fix some documents

mottodora:


docs/source/install.rst
```diff
@@ -42,4 +42,18 @@ You can use ``setup.py`` to install Chainer from the tarball::
...
 
    $ git clone https://github.com/pfnet-research/chainerchem.git
-   $ pip install -e chainerchem
\ No newline at end of file
+   $ pip install -e chainerchem
+
+Run example training code
+========================
+`The official repository <https://github.com/pfnet-research/chainerchem>`_ provides examples
+several graph convolution networks with the Tox21 and QM9 datasets
```
corochann: provides examples
of training several graph convolution networks with the Tox21 and QM9 datasets

docs/source/install.rst
```diff
@@ -42,4 +42,18 @@ You can use ``setup.py`` to install Chainer from the tarball::
...
+(the Tox21 example has inference code as well). You can obtain the code by cloning
+the repository::
```
corochann: The code can be obtained by cloning the repository... is better?

docs/source/install.rst
```diff
@@ -42,4 +42,18 @@ You can use ``setup.py`` to install Chainer from the tarball::
...
    $ git clone https://github.com/pfnet-research/chainerchem.git
-   $ pip install -e chainerchem
\ No newline at end of file
+   $ pip install -e chainerchem
+
+Run example training code
+========================
+`The official repository <https://github.com/pfnet-research/chainerchem>`_ provides examples
+several graph convolution networks with the Tox21 and QM9 datasets
+(the Tox21 example has inference code as well). You can obtain the code by cloning
```
corochann: Just a comment that when we wrote specific current condition of the source code, we might need to modify docs frequently (When new dataset example is added we might need to update this sentence.)

mottodora: I will delete "with the Tox21 and QM9 datasets
+(the Tox21 example has inference code as well)"

PR #32, opened 2017-12-08
Fix example cache

corochann:
Please merget this after PR #25

Related issue: https://github.com/pfnet-research/chainerchem/issues/29#issuecomment-350028510

PR #37, opened 2017-12-14
Add test scripts for examples

delta2323:
This PR adds scripts to test code in the example directory.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -4,9 +4,18 @@ These examples are implemented to train the model.
 
 TODO: Also, trained model is available and one can use only for predicting the value.
 
-## tox21
-12 types of toxity classification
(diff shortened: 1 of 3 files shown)


delta2323:
The known problem of these two scripts is that we cannot run them simultaneously because they share the same cache directories. Also, these scripts remove cache directory every time they start training with different configuration. That means we do not test if we can use cached datasets.

corochann:
If we merge this PR, can we run simultaneously? or it might overwrite the `npz` file which is not expected?
https://github.com/pfnet-research/chainer-chemistry/pull/32

corochann:
I think this PR won't affect to travis CI result, so let me merge it.

PR #46, opened 2017-12-18
[WIP] Save and load pretrained models in Tox21

### LINKED ISSUE #42: Inference code in the Tox21 example does not load trained parameters

delta2323:
This PR fixes #42


examples/tox21/inference_tox21.py
```diff
@@ -44,18 +45,29 @@ def main():
...
 
-    _, _, test = data.load_dataset(method, labels)
+    _, test, _ = data.load_dataset(method, labels)
+    test = test.get_datasets()
+    X_test = D.NumpyTupleDataset(*test[:-1])
+    y_test = test[-1]
 
     predictor_ = predictor.build_predictor(
         method, config['unit_num'], config['conv_layers'], class_num)
+    chainer.serializers.load_npz(os.path.join(args.in_dir, 'snapshot_iter_92'),
```
mottodora: It is kind to be able to specify model by argument. The default value(in this case, 'snapshot_iter_92') should be set by argparse.

delta2323:
Currently we output accuracy of each task in the inference script. But as pointed out in #62, accuracy is not an appropriate metric for imbalanced datasets like Tox21. We need to change more appropriate ones.

Issue #36, opened 2017-12-14
Unittest in test_schnet.py often fails


delta2323:
`test_backward` in `test_schnet.py` checks gradients numerically calculated and one calculated with backpropagation. It often, not everytime, fails both in CPU and GPU.

corochann:
I think this will be blocking issue for merging PR, so can we increase tolerance?

delta2323:
Solved via #47

PR #47, opened 2017-12-19
update tolerance

corochann:
Related to issue https://github.com/pfnet-research/chainer-chemistry/issues/36

Before PR, test_schnet failes once in three times.
After this PR, test_schnet did not fail with 30 times test in my environment.

Since this is blocking issue to merge another PR, I would like it to be merged.


delta2323:
LGTM

PR #52, opened 2017-12-20
Add links

corochann:
Add related links

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,16 +121,23 @@ Links for Chainer Chemistry:
  - Document: [https://chainer-chemistry.readthedocs.io](https://chainer-chemistry.readthedocs.io)
  - Blog: [Release Chainer Chemistry: A library for Deep Learning in Biology and Chemistry](https://preferredresearch.jp/2017/12/18/chainer-chemistry-beta-release/)
 
-Links for other Chainer project:
+Links for other Chainer projects:
 
- - Chainer: A flexible framework of neural networks for deep learning - [chainer/chainer](https://github.com/chainer/chainer)
+ - Chainer: A flexible framework of neural networks for deep learning
+   - Official page: [Website](https://chainer.org/)
+   - Github: [chainer/chainer](https://github.com/chainer/chainer)
  - ChainerRL: Deep reinforcement learning library built on top of Chainer - [chainer/chainerrl](https://github.com/chainer/chainerrl)
  - ChainerCV: A Library for Deep Learning in Computer Vision - [chainer/chainercv](https://github.com/chainer/chainercv)
  - ChainerMN: Scalable distributed deep learning with Chainer - [chainer/chainermn](https://github.com/chainer/chainermn)
  - ChainerUI: User Interface for Chainer - [chainer/chainerui](https://github.com/chainer/chainerui)
- - PaintsChainer: Line drawing colorization using chainer - [pfnet/PaintsChainer](https://github.com/pfnet/PaintsChainer)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,6 +114,26 @@ Use it at your own risk.
...
 Please see the [LICENSE](https://github.com/pfnet-research/chainer-chemistry/blob/master/LICENSE) file for details.
 
+## Links
+
+Links for Chainer Chemistry:
+
+ - Document: [https://chainer-chemistry.readthedocs.io](https://chainer-chemistry.readthedocs.io)
+ - Blog: [Release Chainer Chemistry: A library for Deep Learning in Biology and Chemistry](https://preferredresearch.jp/2017/12/18/chainer-chemistry-beta-release/)
+
+Links for other Chainer project:
```
delta2323: projects

README.md
```diff
@@ -114,6 +114,26 @@ Use it at your own risk.
...
+
+ - Chainer: A flexible framework of neural networks for deep learning - [chainer/chainer](https://github.com/chainer/chainer)
```
delta2323: How about adding a link to titles (Chainer in this case)?

delta2323: Is it better to use the official website (https://chainer.org) as a destination URL?

README.md
```diff
@@ -114,6 +114,26 @@ Use it at your own risk.
...
+ - ChainerRL: Deep reinforcement learning library built on top of Chainer - [chainer/chainerrl](https://github.com/chainer/chainerrl)
+ - ChainerCV: A Library for Deep Learning in Computer Vision - [chainer/chainercv](https://github.com/chainer/chainercv)
+ - ChainerMN: Scalable distributed deep learning with Chainer - [chainer/chainermn](https://github.com/chainer/chainermn)
+ - ChainerUI: User Interface for Chainer - [chainer/chainerui](https://github.com/chainer/chainerui)
+ - PaintsChainer: Line drawing colorization using chainer - [pfnet/PaintsChainer](https://github.com/pfnet/PaintsChainer)
```
delta2323: Which do you think it is better to use the demo URL (https://paintschainer.preferred.tech/) or github repository (https://github.com/pfnet/PaintsChainer) as a destination URL?

README.md
delta2323: It think it is good to add CuPy, too.

README.md
```diff
@@ -114,6 +114,26 @@ Use it at your own risk.
...
+ 
+If you are new with chainer, here is a tutorial to start with:
```
delta2323: If you are new to Chainer, ...

README.md
corochann: noted.

README.md
corochann: > adding a link to titles
What do you mean?

> Is it better to use the official website (https://chainer.org) as a destination URL?
May be we can put both (official page link as well as github link)

README.md
corochann: Maybe we can put both.

README.md
corochann: Noted.

corochann:
Updated.
Is it better to put link on title for Chainer RL/CN/MN/UI?

delta2323:
I imagined something like `[ChainerRL](https://github.com/chainer/chainerrl)` as it is more compact. But current code is also OK as it is a problem of tastes.

corochann:
Then, I would like to use this format for now.
We can consider updating it again any time.

delta2323:
LGTM

PR #57, opened 2017-12-21
Test Python 3.5 and Chainer v2, v4pre in Travis

delta2323:
This PR changes `.travis.yml` to test if Chainer Chemistry work with Chainer v2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,11 +24,11 @@ install:
   - source activate test-environment
   - conda install pip pytest
   - conda install -c rdkit rdkit
-  - if [ "${CHAINER_VERSION}" = "prerelease" ]; then
-      pip install --pre chainer
-    else
-      pip install "${CHAINER_VERSION}"
-    fi
+  - if [ "${CHAINER_VERSION}" = "prerelease" ]; then;
+      pip install --pre chainer;
+    else;
+      pip install "${CHAINER_VERSION}";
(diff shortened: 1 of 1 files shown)


delta2323:
Now the PR is review-ready. 

I change the travis file to test Python 3.5 and the development version of Chainer (v4.0.0b2) as well. Although the number of configurations is increased to 9, it is acceptable as each unit tests are relatively small and fast.

I noticed that the current master branch does not pass unit tests with Chainer v2. I fixed the test code to pass with that version.

corochann:
I feel supporting "prerelease" version of chainer may be a too strong restriction, and we might need to change the test configuration in the future.
However, for now, I think it is good to test these versions to check compatibility so let me merge this.

delta2323:
> I feel supporting "prerelease" version of chainer may be a too strong restriction, and we might need to change the test configuration in the future.

IMO, it is OK that which versions we *test* in Travis and which versions we *support*, or we guarantee that Chainer Chemistry can work with. But I also feel it is not sound that we merge PRs even if some tests  fail, telling authors of PRs to neglect them.

For now, we can ensure Chainer Chemistry works with the prerelease version of Chainer and as it does not take time much time to test all combination in Travis, I think we can benefit more by adding the prerelease version. Let's discuss when we suffer from instability of prerelese versions.

PR #58, opened 2017-12-21
Fix readthedocs setting

### LINKED ISSUE #48: Error in API Reference
It seems that readthedocs fails generating detail of API reference.
http://chainer-chemistry.readthedocs.io/en/latest/dataset.html

mottodora:
Fix #48 
To generate documents properly, we need installing RDKit.
So I wrote some settings for installing RDKit in readthedocs environment.
The result is [here](http://chainer-chemistry-test.readthedocs.io/en/fixdocs/dataset.html)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,6 @@
+name: chainerchem
+type: sphinx
+base: docs/source
+requirements_file: docs/source/requirements.txt
+python:
+  setup_py_install: true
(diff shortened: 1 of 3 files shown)


.readthedocs.yml
```diff
@@ -0,0 +1,8 @@
+name: chainerchem
```
corochann: chainer-chemistry

docs/source/environment.yml
```diff
@@ -0,0 +1,79 @@
+
+name: chainerchem
```
corochann: chainer-chemistry

PR #59, opened 2017-12-21
balanced serial iterator

corochann:

PR #60, opened 2017-12-21
Tox21 balanced example

corochann:
Please merge this after https://github.com/pfnet-research/chainer-chemistry/pull/59

[Note]
1. The option added in tox21 example does not work when user installed library version=0.1.0.

You may test this example by for example,
`python train_tox21.py --label=NR-AR --iterator_type=balanced --method=ggnn -g 0`


examples/tox21/train_tox21.py
```diff
@@ -81,7 +86,20 @@ def main():
     predictor_ = predictor.build_predictor(
         method, args.unit_num, args.conv_layers, class_num)
 
-    train_iter = I.SerialIterator(train, args.batchsize)
+    iterator_type = args.iterator_type
+    if iterator_type == 'serial':
+        train_iter = I.SerialIterator(train, args.batchsize)
+    elif iterator_type == 'balanced':
+        from chainer_chemistry.iterators.balanced_serial_iterator import BalancedSerialIterator  # NOQA
```
delta2323: Pep8 recommends to import modules at the top of files. So, I suggest to move it upward if unless it is unavoidable.

delta2323:
Could you fix the example test scripts to check the Tox21 example works with the balanced serial iterator.

examples/tox21/train_tox21.py
corochann: One note is that when user install library from `pip` command, this import does not work.
Therefore, original option 'serial' is still available in this implementation but 'balanced' is not available from pip library version=0.1.0.

Should I write `try ... except ImportError: print('if you want to use BalancedSerialIterator, please install library from master branch')` or something?

examples/tox21/train_tox21.py
```diff
@@ -51,6 +52,10 @@ def main():
                         default='', help='target label for logistic '
                         'regression. Use all labels if this option '
                         'is not specified.')
+    parser.add_argument('--iterator_type', type=str, choices=iterator_type,
```
delta2323: It is a matter of tastes but I prefer to use hyphens instead of underscores (i.e. `--iterator-type`) because it is visually aligned. `argparse` automatically converts intermediate hyphens into underscores when using this argument as a property. So we can use `args.iterator_type`.

examples/example_test_cpu.sh
```diff
@@ -9,31 +9,36 @@ do
...
     fi
 
     python train_qm9.py --method ${method} --label A --conv_layers 1 --gpu ${gpu} --epoch 1 --unit_num 10
     python train_qm9.py --method ${method} --conv_layers 1 --gpu ${gpu} --epoch 1 --unit_num 10
     cd ../
 done
+
+# BalancedSerialIterator test for Tox21
+cd tox21
+python train_tox21.py --method nfp --label NR-AR --conv-layers 1 --gpu ${gpu} --epoch 1 --unit-num 10 --out nr_ar_nfp_balanced --iterator-type balanced
```
delta2323: Could you add `inference.py` as well to see if the output of `train_tox21.py` is valid even with `BalancedSerialIterator`. Although we know changing type of the iterator does not affect the format of output, it makes the example code more robust for future change.

examples/tox21/train_tox21.py
delta2323: I understood. Then, I want to turn down this suggestion and want you to revert it because the current code show warning about `BalancedSerialIterator` even if users want to use ordinal iterators.

delta2323: As we discussed offline, we will update the minor version when the library code cannot run the example code in the master branch and it is costly or makes the example complicated if we want to support backward compatibility.

corochann:
Added inference tests. I executed example_test_gpu.sh and it was ok.

mottodora:
LGTM!

PR #62, opened 2017-12-21
Report roc auc

corochann:
Please merge this after
https://github.com/pfnet-research/chainer-chemistry/pull/60

I think it is better to show ROC-AUC score instead of binary accuracy in tox21 example.
This PR adds roc-auc report.

However, ROC AUC score is calculated per minibatch, so its value is not accurate... Is it ok to calculate just a "approximate" value of roc auc score during training?
I'd like to get other person's opinion.


corochann:
You can test WeaveNet training with data balancing, and  ROC-AUC score reported during training:

example command:
`python train_tox21.py --label=NR-AR -g 0 --iterator_type=balanced --method=weavenet --eval_mode=1`

delta2323:
> However, ROC AUC score is calculated per minibatch, so its value is not accurate... Is it ok to calculate just a "approximate" value of roc auc score during training?

It seems to me that current implementation calculates ROC and AUC with full datasets. Isn't it true?

corochann:
Yes. I updated implementation.

examples/tox21/roc_auc_evaluator.py
```diff
@@ -0,0 +1,52 @@
...
+        t_total = []
+        for batch in iterator:
+            in_arrays = self.converter(batch, self.device)
+            with chainer.no_backprop_mode(), chainer.using_config('train',
+                                                                  False):
+                y = predictor(*in_arrays[:-1])
+            t = in_arrays[-1]
+            y_data = get_1d_numpy_array(y)
+            t_data = get_1d_numpy_array(t)
+            y_data = y_data[t_data != -1]
```
mottodora: I think this line is mistake.  In tox21 dataset we should ignore data point with -1 label because it means unlabeled (in chainer-chemistry). But ROC-AUC score function can be used for general purpose and scikit learn roc_auc_score can recieve -1 label data (https://github.com/scikit-learn/scikit-learn/blob/a24c8b46/sklearn/metrics/ranking.py#L465-L466). L45 is also.

corochann: I read it, I think it means you can use either of the case

1. positive label is 1 and negative label is 0
2. positive label is 1 and negative label is -1
3. positive label is `pos_label` if specified, all other values are considered as negative labels.

So I think there is no way to "ignore" not to include from scoring.
Thus we need to include it.

chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+        eval_func: Evaluation function called at each iteration.
+        pos_labels (list): labels of the positive class
+        ignore_labels (list): labels to be ignored.
+
+    """
+
+    def __init__(self, iterator, target, predictor=None,
+                 converter=convert.concat_examples,
+                 device=None, eval_hook=None, eval_func=None, name=None,
+                 pos_labels=1, ignore_labels=-1):
```
mottodora: How do you feel adding `neg_label` argument?

chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+    if isinstance(a, (int, float)):
+        return [a, ]
+    else:
+        # expected to be list or some iterable class
+        return a
+
+
+class ROCAUCEvaluator(Evaluator):
+
+    """Evaluator which calculates ROC AUC score
```
mottodora: I would like to add ' this implementation is restricted to the binary classification task'.

chainer_chemistry/training/extensions/roc_auc_evaluator.py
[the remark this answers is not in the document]
corochann: ok

corochann: I wanted similar API with sklearn, which uses only `pos_label` (only one int is avaialble), and rest is considered as negative.

chainer_chemistry/training/extensions/roc_auc_evaluator.py
corochann: ok

corochann:
Updated based on comments & added tests. Could you review?

chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Note that this Evaluator is only applicable to binary classification task.
+
+    Args:
+        iterator: Dataset iterator for the dataset to calculate ROC AUC score.
+            It can also be a dictionary of iterators. If this is just an 
+            iterator, the iterator is registered by the name ``'main'``.
+        target: Link object or a dictionary of links to evaluate. If this is
+            just a link object, the link is registered by the name ``'main'``.
+        converter: Converter function to build input arrays.
```
mottodora: `predictor` is missing?

corochann:
I removed `predictor` argument, it can be specified by `eval_func` instead.
Is it ok?

mottodora:
LGTM!

PR #65, opened 2017-12-25
Fix documents

mottodora:
fix some docstring to be able to build without warning.
(Current codes produce many warinings.


delta2323:
As you can see in the current [document](http://chainer-chemistry.readthedocs.io/en/latest/generated/chainer_chemistry.dataset.preprocessors.construct_atomic_number_array.html#chainer_chemistry.dataset.preprocessors.construct_atomic_number_array), some generated documents, especially argument lists collapse. So, we first need to fix them.

It seems we need to add a line break after colons that separates an argument name and its description, although current format works in Chainer.

PR #69, opened 2017-12-27
[ToBeClosed][WIP] Chainer sklearn qm9

corochann:


corochann:
I think `Classifier` and `Regressor` are introduced now, so inference can be done in very simple way.

Only the remaining advantage of this PR is training by `fit`.
But maybe we can consider different way to support wasy training? it may be ok to just close this PR.

delta2323:
@corochann Can we close this PR?

corochann:
ok

PR #75, opened 2018-01-02
Include napoleon as an extension

delta2323:
This PR adds [napoleon](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#) to `conf.py`. It is a Sphinx extension that enables us to write documents following Google Python Style Guide. As some of our documents are written in this style, we need to include it.


delta2323:
This change removes most of warnings in building the document. So, I guess we need not to apply changes in #65.

corochann:
I think it is ok to merge this.
@mottodora Could you review?

PR #77, opened 2018-01-03
Add documents to concat_mols

delta2323:
This PR adds a document to `concat_mols`

### DIFF
--- a/chainer_chemistry/dataset/converters.py
+++ b/chainer_chemistry/dataset/converters.py
@@ -2,7 +2,36 @@ import chainer
 
 
 def concat_mols(batch, device=None, padding=0):
-    """Concatenates a list of mols into array(s).
+    """Concatenates a list of molecules into array(s).
+
+    This function converts "an arrays of tuples" into "a tuple of arrays".
+    Given an list of examples each of which consists of
+    multiple arrays, this function takes an array in the same "relative"
+    position within an example from each example, concatenates them along the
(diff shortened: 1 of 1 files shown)


chainer_chemistry/dataset/converters.py
```diff
@@ -2,7 +2,36 @@
...
+    Then, this function concatenates ``x`` 's into one array, and ``y`` 's
+    into another array, and returns a tuple of these two arrays. Another
+    example: consider each example is a dictionary of two entries whose keys
+    are ``'x'`` and ``'y'``, respectively, and values are arrays. Then, this
+    function concatenates ``x`` 's into one array, and ``y`` 's into another
+    array, and returns a dictionary with two entries ``x`` and ``y`` whose
+    values are the concatenated arrays.
+
+    When the arrays to concatenate have different shapes, the behavior depends
+    on the ``padding`` value. If ``padding`` is ``None`` (default), it raises
```
corochann: I guess you copied from 
https://github.com/chainer/chainer/blob/master/chainer/dataset/convert.py


`padding=0` is default, this sentence is confusing.

corochann:
And if possible, I guess adding an example is more easier to understand the behavior of this function, as written in 
https://github.com/chainer/chainer/blob/master/chainer/dataset/convert.py

delta2323:
OK, I'll do that.

corochann:
Let me take over the remaining work.

PR #79, opened 2018-01-16
return smiles as numpy array

corochann:
Originally smiles is returned as `list` with `get_qm9` or `get_tox21` method.

However this PR is to change its type to return `numpy array`.

The motivation is that since dataset features are of numpy array, I felt it is more unified when numpy array is returned.
Also, advanced indexing can be used when the type is numpy array instead of `list`

However since `list` type was originally returned, **this PR breaks backward compatibility** while I guess the effect is small (because there is no example code to use `retain_smiles=True` for now.


mottodora:
LGTM!

Issue #64, opened 2017-12-25
Add unittest: balanced serial iterator's serialize method


corochann:

corochann:
Done by PR #86 
Thank you @n-yoshikawa

PR #86, opened 2018-02-04
Add test for balanced serial iterator

n-yoshikawa:
This PR solves issue #64

### DIFF
--- a/tests/iterators_tests/test_balanced_serial_iterator.py
+++ b/tests/iterators_tests/test_balanced_serial_iterator.py
@@ -4,23 +4,22 @@ import pytest
 from chainer import serializer
 
 from chainer_chemistry.datasets.numpy_tuple_dataset import NumpyTupleDataset
-
 from chainer_chemistry.iterators.balanced_serial_iterator import BalancedSerialIterator  # NOQA
 from chainer_chemistry.iterators.balanced_serial_iterator import IndexIterator  # NOQA
 
 
 class DummySerializer(serializer.Serializer):
 
-    def __init__(self, target):
-        super(DummySerializer, self).__init__()
-        self.target = target
+    def __init__(self, path=[]):
+        self.path = path
+        self.called = []
 
     def __getitem__(self, key):
-        raise NotImplementedError
+        return DummySerializer(self.path + [key])
 
     def __call__(self, key, value):
-        self.target[key] = value
-        return self.target[key]
+        self.called.append((key, value))
+        return value
 
 
 def test_index_iterator():
@@ -55,8 +54,7 @@ def _test_index_iterator_no_shuffle():
     assert indices3[0] == index_list[1]
     assert indices3[1] == index_list[2]
 
-    target = dict()
-    ii.serialize(DummySerializer(target))
+    ii.serialize(DummySerializer())
     assert isinstance(ii.current_index_list, numpy.ndarray)
     assert len(ii.current_index_list) == len(index_list)
     assert numpy.array_equal(ii.current_index_list, numpy.asarray(index_list))
@@ -82,8 +80,7 @@ def _test_index_iterator_with_shuffle():
         for index in indices:
             assert index in index_list
 
-    target = dict()
-    ii.serialize(DummySerializer(target))
+    ii.serialize(DummySerializer())
     for index in ii.current_index_list:
         assert index in index_list
     assert ii.current_pos == (3 + 6) % len(index_list) + 2
@@ -118,9 +115,13 @@ def _test_balanced_serial_iterator_no_batch_balancing():
     assert numpy.sum(labels_batch == 1) == 3
     assert numpy.sum(labels_batch == 2) == 3
 
-    # This does not work!
-    # target = dict()
-    # iterator.serialize(DummySerializer(target))
+    iterator.serialize(DummySerializer())
+    assert iterator.current_position == 0
+    assert iterator.epoch == 1
+    assert iterator.is_new_epoch == True
+    for order in iterator._order:
+        assert order <= 9
+    assert iterator._previous_epoch_detail == 0.0
 
 
 def _test_balanced_serial_iterator_with_batch_balancing():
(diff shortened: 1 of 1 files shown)


delta2323:
Great! Could you remove [WIP] when the PR is review-ready?

tests/iterators_tests/test_balanced_serial_iterator.py
```diff
@@ -1,11 +1,27 @@
...
+from chainer import serializer
+
+from chainer_chemistry.datasets.numpy_tuple_dataset import NumpyTupleDataset
 from chainer_chemistry.iterators.balanced_serial_iterator import BalancedSerialIterator  # NOQA
 from chainer_chemistry.iterators.balanced_serial_iterator import IndexIterator  # NOQA
 
 
+class DummySerializer(serializer.Serializer):
+
+    def __init__(self, path=[]):
```
delta2323: Use `()`, which is immutable as a default value rather than `[]`.  See e.g. [here](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html).

tests/iterators_tests/test_balanced_serial_iterator.py
```diff
@@ -38,6 +54,12 @@ def _test_index_iterator_no_shuffle():
     assert indices3[0] == index_list[1]
     assert indices3[1] == index_list[2]
```
delta2323: I think it would be better to separate serialization tests from `_test_index_iterator_no_shuffle` to make each do one thing. IMO, it is acceptable that this change occurs code duplication to some extent. For example how about making `test_index_iterator_serialization` that runs `_test_index_iterator_serialization_no_shuffle` and `_test_index_iterator_serialization_with_shuffle` (function names are tentative).

tests/iterators_tests/test_balanced_serial_iterator.py
```diff
@@ -88,6 +115,14 @@ def _test_balanced_serial_iterator_no_batch_balancing():
     assert numpy.sum(labels_batch == 1) == 3
     assert numpy.sum(labels_batch == 2) == 3
 
+    iterator.serialize(DummySerializer())
+    assert iterator.current_position == 0
+    assert iterator.epoch == 1
+    assert iterator.is_new_epoch == True
+    for order in iterator._order:
+        assert order <= 9
+    assert iterator._previous_epoch_detail == 0.0
```
delta2323: You do not have to check the value of private properties as its specification is not documented. Ideally, we should write tests with documented APIs.

corochann:
Do you want to test for "only serialization" or both "serialization -> deserialization"?

For example,
1. Chainer's iterator test is checking that `iterator` can be saved and loaded by  "serialization -> deserialization". And loaded instance has same attribute value with saved instance.
https://github.com/chainer/chainer/blob/master/tests/chainer_tests/iterators_tests/test_serial_iterator.py#L251-L255

2. Chainer's updater test is checking only "serialization" by preparing mock `DummyIterator` and `DummyOptimizer` and check `serialize_called` attribute to see serialization is done expectedly.

https://github.com/chainer/chainer/blob/master/tests/chainer_tests/training_tests/updaters_tests/test_standard_updater.py


For this PR, I think we can take first way (check "serialization -> deserialization") by defining `DummyDeserializer` as well.

tests/iterators_tests/test_balanced_serial_iterator.py
```diff
@@ -101,14 +187,89 @@ def _test_balanced_serial_iterator_with_batch_balancing():
...
+    x = numpy.arange(8)
+    t = numpy.asarray([0, 0, -1, 1, 1, 2, -1, 1])
+    iterator = BalancedSerialIterator(NumpyTupleDataset(x, t), batch_size=9,
+                                      labels=t, ignore_labels=-1,
+                                      batch_balancing=False)
+    batch = iterator.next()
+
+    assert iterator.current_position == 0
+    assert iterator.epoch == 1
+    assert iterator.is_new_epoch == True
```
delta2323: `assert iterator.is_new_epoch` is enough.

tests/iterators_tests/test_balanced_serial_iterator.py
```diff
@@ -101,14 +187,89 @@ def _test_balanced_serial_iterator_with_batch_balancing():
...
+        current_index_list_orig[ii_label] = index_iterator.current_index_list
+        current_pos_orig[ii_label] = index_iterator.current_pos
+
+    iterator = BalancedSerialIterator(NumpyTupleDataset(x, t), batch_size=9,
+                                      labels=t, ignore_labels=-1,
+                                      batch_balancing=False)
+    iterator.serialize(DummyDeserializer(target))
+    assert iterator.current_position == 0
+    assert iterator.epoch == 1
+    assert iterator.is_new_epoch == True
```
delta2323: ditto

tests/iterators_tests/test_balanced_serial_iterator.py
```diff
@@ -101,14 +187,89 @@ def _test_balanced_serial_iterator_with_batch_balancing():
...
+    iterator = BalancedSerialIterator(NumpyTupleDataset(x, t), batch_size=3,
+                                      labels=t, ignore_labels=-1,
+                                      batch_balancing=True)
+    batch1 = iterator.next()
+    batch2 = iterator.next()
+    batch3 = iterator.next()
+
+    assert iterator.current_position == 0
+    assert iterator.epoch == 1
+    assert iterator.is_new_epoch == True
```
delta2323: ditto

tests/iterators_tests/test_balanced_serial_iterator.py
```diff
@@ -101,14 +187,89 @@ def _test_balanced_serial_iterator_with_batch_balancing():
...
+        current_index_list_orig[ii_label] = index_iterator.current_index_list
+        current_pos_orig[ii_label] = index_iterator.current_pos
+
+    iterator = BalancedSerialIterator(NumpyTupleDataset(x, t), batch_size=3,
+                                      labels=t, ignore_labels=-1,
+                                      batch_balancing=True)
+    iterator.serialize(DummyDeserializer(target))
+    assert iterator.current_position == 0
+    assert iterator.epoch == 1
+    assert iterator.is_new_epoch == True
```
delta2323: ditto

delta2323:
LGTM

PR #89, opened 2018-02-14
Support GCN

anaruse:
This PR aims to add a GCN model and support batch normalization for graph convolution. Regarding GCN model, please refer to https://arxiv.org/abs/1609.02907, though the GCN model in this branch is bit different from the model in the paper. I also modified example of tox21, so that you can test the GCN model with tox21 dataset.


corochann:
Thank you very much for the PR!
Let me have time to read the paper briefly and review it.

This is question so far.
1. Are `graph_convolution` functions and `link` is used? It is imported but seems not used.
2. My concern is "namespace" issue. I think GCN (GraphConvolutionalNetwork) is very "general" term compared to other network implementation, do you have any comment on this?

anaruse:
Thank you for your feedback!

Re 1., I'm sorry for that I forgot to remove them. They are not used now. will remove them.

Re 2., I have the same opinion that the name of "GCN" is too general, though the authors calls it just "GCN" in the paper. I once thought that "LBM" might be a good name for it because there is a paper ([link](https://arxiv.org/abs/1704.01212)) that classifies this model as one of "Laplacian Based Methods". However, "LBM" is bit confusing to me, mainly because "LBM" means "Lattice Boltzmann Method" to me...

corochann:
Re 2., 
Actually I'm considering importing graph cnn at
https://github.com/pfnet-research/chainer-graph-cnn
and same issue happens that the name is too general. 

I'm currently considering to use the name "Spectral"GraphConvolution for this.
https://github.com/corochann/chainer-chemistry/blob/spectral_gcnn/chainer_chemistry/models/spectral_graph_cnn.py#L15-L19

Let us consider naming issue with other members as well.

chainer_chemistry/models/gcn.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __call__(self, x):
+        # --- Readout part ---
+        h = self.output_weight(x)
+
+        #
+        h = functions.softmax(h, axis=2)  # softmax along channel axis
+        y = functions.sum(h, axis=1)  # sum along node axis
+        return y
+
```
corochann: Original paper does not define "Readout" (get feature of "whole graph"), instead, their output is for each node by simply taking softmax only.

So, `output_weight` and `sum` is added by you to extract "graph feature"?
Is my understanding correct?

chainer_chemistry/models/gcn.py
```diff
@@ -0,0 +1,132 @@
...
+        if isinstance(adj, Variable):
+            w_adj = adj.data
+        else:
+            w_adj = adj
+        w_adj = Variable(w_adj, requires_grad=False)
+        
+        for i, (gconv, bnorm) in enumerate(zip(self.gconvs,
+                                               self.bnorms)):
+            h = gconv(h, w_adj)
+            h = bnorm(h)
```
corochann: I referred pytorch implementation by author.
https://github.com/tkipf/pygcn/blob/master/pygcn/models.py#L14

batch norm is originally not used, so is it added by you?
If so, can you add an option in `__init__` kwargs to use/not use batch normalization?

examples/tox21/run_gcn.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -x
```
corochann: Is this file necessary?

examples/tox21/train_tox21.py
```diff
@@ -36,6 +36,8 @@
 import data
 import predictor
 
+import numpy
```
corochann: Is this line necessary? seems numpy is not used in change.

corochann:
I read the paper, https://arxiv.org/pdf/1609.02907.pdf
and looked your code. Overall it seems nice.
I just put some comment.

Re 2., 
In conclusion, it is written as
> Our GCN model uses an efficient layer-wise propagation rule that is based on a **first-order approximation**
> **of spectral convolutions on graphs**.

and in Table 3, they finally adopt *Renormalized* version of the Eq (8).

So in my opinion this is "Renormalized First order Spectral Graph Convolutional Network".
However this name is of course long, how about `RenormalizedSpectralGraphConvolutionalNetwork` (`RSGCN`) ?

delta2323:
Regarding the naming issue. I would suggest how others call the method. For example:

* Is there any paper that refers GCN?
* Does someone mention and discuss the paper at e.g. Stack Overflow?
* Also, I'm wondering if other frameworks (e.g. DeepChem) implement the paper.

If there isn't. We might need to coin some word.

corochann:
As far as I know, there are several graph-conv methods which uses Laplacian matrix formulation.

[1]. 'Convolutional Neural Networks on Graphs with Fast Localized Spectral Filtering' https://arxiv.org/abs/1606.09375

[2]. This PR. SEMI-SUPERVISED CLASSIFICATION WITH GRAPH CONVOLUTIONAL NETWORKS
https://arxiv.org/pdf/1609.02907.pdf

And these are usually referred as grouped way, e.g.,
A. https://arxiv.org/abs/1704.01212 referred them as "Laplacian based methods"
B. https://arxiv.org/pdf/1801.03226.pdf referred them as "Spectral Graph Convolution"

That is why I suggested 
`RenormalizedSpectralGraphConvolutionalNetwork` (`RSGCN)`
to use "SpectralGraphConvolution" plus the word "Renormalized" to differenciate [1] and [2]

chainer_chemistry/models/gcn.py
anaruse: Yes.  To be more precise, I had no idea what to do with readout, so I used the NFP's readout implementation.

chainer_chemistry/models/gcn.py
anaruse: OK, will add an option to select whether to use bath normalization or not.

examples/tox21/run_gcn.sh
anaruse: No... I thought I removed it...

anaruse:
Thank you for your feedback!  I've fixed the branch based on your suggestion. Please check again!

- Name issue: RSGCN looks good to me (I think it is better than LBM). 
- Support sparse matmul: This will be needed specifically when graph size is large.

corochann:
Thank you for fix. I'd like to merge it after following discussion.

1. Naming issue, let me know your opinion if it is ok with `RenormalizedSpectralGraphConvolutionalNetwork` or not. @delta2323 

I also briefly checked deepchem, molecule.ai but I could not find any existing naming for this.
http://moleculenet.ai/models
 
2. ReadOut module.
If there is no original paper suggestion, I would like to take `readout` as `init` args of the Network as well, so that user can change if they want.
Default value of `readout` may be `None` and in that case we can use your implementation of `GCNReadout` (but I want to add a docstring that this is not the original paper's implementation).
What do you think? 
(I am personally thinking that just `sum` along atom axis is enough and extra `output_weight` and `softmax` is not necessary. Any user may try these kind of different readout module by passing as args.)

anaruse:
Thank you for your feedback, @corochann !

I'm not particular about readout, so I added `readout` option to the model so that users can specify what function to use for readout. It uses "sum readout" when no function is specified by users.

delta2323:
Regarding the naming issue, I looked up several papers that refer the original paper and how they call their method.

(1) Papers that only refer the original paper and do not name the algorithm: [URL](https://arxiv.org/abs/1704.06857), [URL](https://arxiv.org/abs/1702.00539).
(2) Papers that call the algorithm "GCN" [URL](https://papers.nips.cc/paper/6703-inductive-representation-learning-on-large-graphs) [URL](https://arxiv.org/abs/1611.07308).

(3) A paper that calls both the algorithm and other ones "GCN": [URL](https://arxiv.org/pdf/1703.06103.pdf).

As we expect, it seems that there are no de-facto standard of the name except GCN and that GCN can be also used in more general meaning. Therefore, I agree with putting   `RenormalizedSpectralGraphConvolutionalNetwork` or `RSGCN` in short.

corochann:
@anaruse 
Could you change the name from `GCN` to `RSGCN`?
And if possible could you write a docstring that its name `Renormalized Spectral Graph Conlutional Network` is named by us and it is not original authors naming.

After above change, I would like to test the model performance to see if training progresses in my environment and merge it.

anaruse:
I've changed the name from GCN to RSGCN and added some docs. Please review again!

corochann:
Thank you for the change, LGTM.

When I test the model performance with ROC AUC Evaluator (https://github.com/pfnet-research/chainer-chemistry/pull/62), I found the following.

1. It is better to add `softmax` function in default readout func.
The model train progress becomes much stable (somehow) with adding softmax function in readout as follows.

```
def _readout_sum(x):
    h = functions.softmax(x, axis=2)  # softmax along channel axis
    y = functions.sum(h, axis=1)  # sum along node axis
    return y
```

2. Support a case when `readout` is a class of `Link`.
Currently `readout` args in `__init__` is not registered inside `init_scope`. 
But user may want to pass `Chain` class as `readout`, it becomes more flexible to check if `readout` isinstanceof Link or not and register inside `init_scope` if so.


**It is ok for me to send above PR 1 & 2 after merge this PR at first.** Is it ok for you?

anaruse:
Thank you for merging the PR!

Regarding change 1. and 2., I'm fine with those changes. Thanks.

PR #90, opened 2018-02-19
Sparse matmul

anaruse:
This PR aims to support sparse matmul for graph convolution network.

A function named `sparse_matmul` is added with this PR. The use case of this function is "C (dense) = A (sparse) * B (dense)" and you can use it as follows:

    c = F.sparse_matmul(b, a_value, a_row_index, a_col_index)

A sparse matrix storage format adopted by this PR is simple one which keeps "value", "row index" and "column index" of sparse matrix separately. This is mainly because it is easy to implement custom cupy kernel for this format and you can use the kernel both for forward and backward. However, I have to say that I have no idea on efficient numpy implementation of sparse matmul with this format. In other word, you can use this function only on GPU now. I would like to ask numpy magician regarding the implementation with numpy.

Furthermore, sparse version of GCN is implemented as an use case of sparse matmul. So, you can test it with tox21 dataset. Any feedback would be appreciated.


corochann:
Thank you for your quick implementation.
I guess your implementation follows the "COO" sparse matrix format.
I  guess `scipy.sparse` is usually used for CPU version of sparse matrix handling. But it does not support 3-dim sparse matrix which includes `minibatch` axis, so naive extension seems difficult.

Let us consider & review how to deal with sparse matrix implementation for graph convolution.

One question: How is the speed difference between `GCN` and `SparseGCN`?
Is it faster to use `SparseGCN`??

corochann:
I was also investigating sparse matrix, and testing performance of `cupy.sparse`.
Let me just share it.

```
"""
http://ohke.hateblo.jp/entry/2018/01/07/230000
https://docs-cupy.chainer.org/en/stable/reference/sparse.html


Please install library
pip install git+https://github.com/corochann/chainerex
"""
# import numpy as xp
# from scipy import sparse
import numpy as np
import cupy as xp
from cupy import sparse
from chainer import cuda


def allclose(a, b):
    return np.allclose(cuda.to_cpu(a), cuda.to_cpu(b))


# 1000×1000の疎行列の作成
# dimensions = 50000  # OutOfMemory error
dimensions = 10000
# dimensions = 10000
# dimensions = 1000
# dimensions = 100
# sparse_matrix = xp.zeros((dimensions, dimensions), dtype=xp.int64)
sparse_matrix = xp.zeros((dimensions, dimensions), dtype=xp.float32)

# non_zero_elements = 10000  # 非ゼロ要素は全体の1%
# non_zero_elements = 1000  # 非ゼロ要素は全体の0.1%
non_zero_elements = 100  # 非ゼロ要素は全体の0.01%
count = 0
rows = xp.random.randint(0, dimensions, 20000)
columns = xp.random.randint(0, dimensions, 20000)
values = xp.random.randint(0, 10, 20000)
for i in range(0, 20000):
    if sparse_matrix[rows[i], columns[i]] == 0:
        sparse_matrix[rows[i], columns[i]] = values[i]
        count += 1
        if count == non_zero_elements:
            break

# sparse_lil = sparse.lil_matrix(sparse_matrix)
sparse_csr = sparse.csr_matrix(sparse_matrix)


import chainerex.utils as cl
tm = cl.TimeMeasure()
# 疎行列の内積速度の計測

tm.update('start')
a = xp.dot(sparse_matrix, sparse_matrix)
tm.update('xp dot xp-> {} {}'.format(type(a), a.shape))
# b = sparse_lil * sparse_lil
# tm.update('lil dot -> {} {}'.format(type(b), b.shape))
c = sparse_csr * sparse_csr
tm.update('csr * csr -> {} {}'.format(type(c), c.shape))
cdot = sparse_csr.dot(sparse_csr)
tm.update('csr dot csr -> {} {}'.format(type(cdot), cdot.shape))
d = sparse_csr.dot(sparse_matrix)
tm.update('csr dot xp -> {} {}'.format(type(d), d.shape))
e = sparse_csr.dot(sparse_matrix.T)
tm.update('csr dot xp.T -> {} {}'.format(type(e), e.shape))
```

its result is:
```
$ python sparse_matrix_dot_time_cupy.py
[TimeMeasure] start: 0.000002 sec
[TimeMeasure] xp dot xp -> <class 'cupy.core.core.ndarray'> (10000, 10000): 0.000242 sec
[TimeMeasure] csr * csr -> <class 'cupy.sparse.csr.csr_matrix'> (10000, 10000): 0.225117 sec
[TimeMeasure] csr dot csr -> <class 'cupy.sparse.csr.csr_matrix'> (10000, 10000): 0.000281 sec
[TimeMeasure] csr dot xp -> <class 'cupy.core.core.ndarray'> (10000, 10000): 0.002072 sec
[TimeMeasure] csr dot xp.T -> <class 'cupy.core.core.ndarray'> (10000, 10000): 0.000165 sec
```

It shows that the improvement is not so huge (0.000242 sec -> 0.000165 sec), and only when memory alignment is optimal (e.g., csr dot xp is slower than xp dot xp).

corochann:
I was considering a way to utilize `cupy.sparse` so that we can get benefit of `cuSPARSE`.
In that case, we may need to use csr format and it seems not easy to use `matmul`.
Only `dot` operation is supported in sparse matrix.

But when I look at the picture uploaded on here,
https://github.com/tkipf/gcn#graph-classification
it seems it is possible to use sparse matrix by concatenating along minibatch axis to represent nodes (atoms) of each molecules.

anaruse:
Thank you for your comment.

Re: performance of `GCN` and `SparseGCN`, I think we should measure the performance using various size of data-sets. So, I made some synthetic data-sets named `Tox21xN` in which number of atoms of each molecule is N times as many as original Tox21. The measured results are as follows. 

    [Elapsed time (sec)]
                GCN     SparseGCN
    Tox21       41.4       37.2
    Tox21x2     46.7       39.5
    Tox21x4     91.9       41.5
    Tox21x8    274         41.4
    Tox21x16     NA        45.6
    Tox21x32     NA        51.6
    Tox21x64     NA        75.7
    Tox21x128    NA       125

    Batch size: 128
    Epochs: 20
    Chainer: 3.3.0
    CuPy: 2.3.0
    CUDA: 9.0
    cuDNN: 7.0
    Python: 3.6.1 (pyenv anaconda3-4.4.0)
    GPU: Tesla V100
    CPU: Intel Xeon E5-2698

The result says that sparse matmul is kind of "must to have" when graph size is big.

corochann:
Thank you for experiment & quick response.
It seems quite promising, we should consider sparse matmul functionality.
Let us discuss about CPU version implementation & how to deal with API.

anaruse:
I've updated the branch.
- The latest master (a5e8f9e70627ae733b45749f92296b0cb67884a4) is based to reflect some changes.
- `sparse_matmul` implementation submitted to Chainer (https://github.com/chainer/chainer/pull/4397) is adopted.
- The name of model "SparseGCN" is changed to "SparseRSGCN".
- The option `padding` in `concat_mol` is appropriately set when "SarseRSGCN" is used.

anaruse:
Now, `sparse_matmul` is available on CPU too.

corochann:
I have an idea of efficient sparse matmul computation on CPU.
Currently I guess you iterate over minibatch axis, since sparse matmul is only applicable to 2-dim matrix.
https://github.com/pfnet-research/chainer-chemistry/pull/90/files#diff-3a3f705484aa44f5789ea2dcb039de9eR116

Then, how about creating a big sparse coo matrix whose `row_index` & `col_index` is shifted according to the batch index?
Please see this figure for intuitive understanding.
https://user-images.githubusercontent.com/7347296/34198790-eb5bec96-e56b-11e7-90d5-157800e042de.png

So let's say you want to calculate `F.sparse_matmul(B (sparse), A (dense))`, we can reshape `A` to the following, and reconstruct sparse matrix `B'` in following way,
A (minibatch, num_nodes, ch) --> A' (minibatch * atoms, ch)
B (minibatch, num_nodes, num_nodes) --> B' (minibatch * atoms, minibatch * atoms)

Then we can calculate to obtain
`C' = F.sparse_matmul(B', A')` to obtain `C' (minibatch * num_nodes, ch)`, finally we can reshape it to obtain `C (minibatch, num_nodes, ch)`.
Since `B'` is a big sparse matrix, we only need to calculate `sparse_matmul` once.

What do you think of this idea (I am not testing the function, I may be saying incorrect or impossible idea...)?

anaruse:
Basically, I think it is good idea to concatenate many small sparse matrices together and make a single big sparse matrix and run `sparse_matmul` once. If the sparse matrices concatenation is fast enough compared with computation of `sparse_matmul`, it will be good not only for CPU but also for GPU. Probably, it is also good for computation of gradients of dense matrix (more exactly, batch of dense matrices) because the single big sparse matrix can be re-used for that computation (when it is COO format).

My concern now is how to compute gradients of sparse matrices. It needs to perform "g_A (sparse) = g_C (dense) * B (dense)" (assuming you perform "C (dense) = A (sparse) * B (dense)" in forward path). For GPU, I wrote custom kernel with CuPy for this computation. But, for CPU, there is no functions in SciPy and Numpy to perform this computation, AFAIK. Thus, I first perform "**g_A (dense)** = g_C (dense) * B (dense)" using `matmul` and then make "**g_A (sparse)**" by picking up necessary elements from "g_A (dense)"
(please refer to https://github.com/anaruse/chainer-chemistry/blob/fc3f184dc589564197578834721e5d0498d0b194/chainer_chemistry/functions/sparse_matmul.py#L310). As you can see, we can not take this approach for single big sparse matrix, because shape of "g_A (dense)" becomes `(minibatch * num_nodes, minibatch * num_nodes)` and it is too large. Maybe we can iterate over batch axis for this, but it means we have to keep two types of sparse matrices: batched sparse matrix and concatenated sparse matrix. It seems to me that it is not excellent solution.

What would you think on this?

BTW, computation of gradients of sparse matrices is not necessary for RSGCN. If we only think about networks like RSGCN, we can take "sparse matrices concatenation" approach.

corochann:
Ah, I see... situation is different between `forward` and `backward`...

> If we only think about networks like RSGCN, we can take "sparse matrices concatenation" approach.
I also understand that we may use different approach when "only forward" is necessary or "forward & backward" is necessary for optimization...

So far, for graph convolution network, I think we only need "forward" computation for sparse matrix `A` and we only need gradient for dense matrix `B` most of the time.

anaruse:
Although it might have looked as if I was rather negative to your idea, I think it is worth a try.  I will first implement it in the way that it concatenates sparse matrices on demand basis.

corochann:
Thank you very much for discussion. I think we should decide our direction/implementation design.
So sorry but please let us have time to consider/discuss. Please prioritize your other work until then.

corochann:
Sorry for the review delayed.
We would like to proceed review & merge (maybe as experimental feature as first stage).
If you have updates on your code, please let us know.

anaruse:
For now, I've updated the latest code on the branch. The following features were added.
* Fused sparse matmul:  You can test this feature using option `--flatten` in tox21 example. Note that you need to use sparse_rsgcn model to test this feature. When you select this mode, multiple sparse matrices are first to be fused on CPU then transferred to GPU. Please see `concat_sparse_rsgcn` for details.
* Tox21xN dataset:  This is only for performance measurement. you can specify N with option `--multiplier` in tox21 example.

anaruse:
Several conflicts to master branch have been resolved.

corochann:
I looked through your code, currently following implementations are done

・Sparse matrix Calculation
１．sparse_matmul with fuse
２．sparse_matmul with Minibatch axis
in different kernel

・converter for Fuse sparse matrix
・Artificial big Tox21 dataset
・SparseRSGCN implementation

Since there are many things to review, we would like to merge it as new branch `experimental_sparse` to brush up & discuss implementation. Is this direction ok for you? @anaruse

anaruse:
Yes, I agree with you about how to proceed the review.

corochann:
Let me merge this PR on `experimental_sparse` branch.

Issue #82, opened 2018-02-02
add scikit-learn dependency


mottodora:

corochann:
Also add scipy dependency, since old version of sklearn does not install it automatically.

corochann:
https://github.com/pfnet-research/chainer-chemistry/pull/93

PR #93, opened 2018-02-22
Add scipy in dependency

corochann:
Issue: https://github.com/pfnet-research/chainer-chemistry/issues/82

scikit-learn does not install scipy automatically, so we explicitly include scipy as dependency.


mottodora:
LGTM!
I checked

PR #94, opened 2018-02-22
Return dict that contains dataset and smiles from parser.parse

### LINKED ISSUE #92: return smiles from parser.parse
It is better to return smiles by `parser.parse` function instead of internally keep `smiles` attribute.

mottodora:
resolve #92 
My main concern is default argument. 
For now, I set True as `return_smiles` default argument.
If you have opinion, please tell me.

### DIFF
--- a/chainer_chemistry/dataset/parsers/csv_file_parser.py
+++ b/chainer_chemistry/dataset/parsers/csv_file_parser.py
@@ -164,8 +164,8 @@ class CSVFileParser(BaseFileParser):
         if isinstance(result, tuple):
             if self.postprocess_fn is not None:
                 result = self.postprocess_fn(*result)
-            return NumpyTupleDataset(*result), smiles_list
+            return NumpyTupleDataset(*result), numpy.array(smiles_list)
         else:
             if self.postprocess_fn is not None:
                 result = self.postprocess_fn(result)
-            return NumpyTupleDataset(result), smiles_list
+            return NumpyTupleDataset(result), numpy.array(smiles_list)
--- a/chainer_chemistry/dataset/parsers/csv_file_parser.py
+++ b/chainer_chemistry/dataset/parsers/csv_file_parser.py
@@ -1,4 +1,3 @@
-from collections import namedtuple
 from logging import getLogger
 
 import numpy
@@ -57,7 +56,7 @@ class CSVFileParser(BaseFileParser):
                 If set to False, this function returns preprocessed dataset and
                 `None`.
 
-        Returns (namedtuple): Dataset, 1-d numpy array with
+        Returns (dict): dictionary that contains Dataset, 1-d numpy array with
             dtype=object(string) which is a vector of smiles for each example
             or None.
 
@@ -164,13 +163,15 @@ class CSVFileParser(BaseFileParser):
             result = pp.process(filepath)
 
         smileses = numpy.array(smiles_list) if return_smiles else None
-        Result = namedtuple("ParseResult", 'dataset,smiles')
+        parse_result = {}
 
         if isinstance(result, tuple):
             if self.postprocess_fn is not None:
                 result = self.postprocess_fn(*result)
-            return Result(NumpyTupleDataset(*result), smileses)
+            parse_result.update({"dataset": result, "smiles": smileses})
+            return parse_result
         else:
             if self.postprocess_fn is not None:
                 result = self.postprocess_fn(result)
-            return Result(NumpyTupleDataset(result), smileses)
+            parse_result.update({"dataset": result, "smiles": smileses})
+            return parse_result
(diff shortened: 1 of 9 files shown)


chainer_chemistry/dataset/parsers/sdf_file_parser.py
```diff
@@ -129,24 +130,12 @@ def parse(self, filepath, retain_smiles=False):
             # Spec not finalized yet for general case
             result = pp.process(filepath)
 
+        smiles_list = numpy.array(smiles_list) if return_smiles else None
         if isinstance(result, tuple):
             if self.postprocess_fn is not None:
                 result = self.postprocess_fn(*result)
-            return NumpyTupleDataset(*result)
+            return NumpyTupleDataset(*result), smiles_list
```
delta2323: It seems weird to me to return `None` as a second element. Apparently it has several drawbacks, for example, because (1) it breaks compatibility and (2) as users know the second element is always `None` when `return_smiles` is `False`, it is meaningless to them. But on the other hand, if we remove this `None`, users need branching in user code depending on the value of `return_smiles`. It seems to me to consider more what is expected specification of parsers.

chainer_chemistry/dataset/parsers/csv_file_parser.py
```diff
@@ -107,9 +108,9 @@ def parse(self, filepath, retain_smiles=False):
                     if self.postprocess_label is not None:
                         labels = self.postprocess_label(labels)
 
-                    if retain_smiles:
+                    if return_smiles:
                         assert standardized_smiles == Chem.MolToSmiles(mol)
-                        self.smiles.append(standardized_smiles)
+                        smiles_list.append(standardized_smiles)
```
delta2323: I would like to confirm the motivation for the `return_smiles` option. I thought the typical use case of this option is to turn it off when users want to reduce memory consumption, as we can expect computation time does not decrease so much, as appending to a list does not require take much time and it does create `smiles` anyway even if `return_smiles` is `False`.

corochann: Most of the time, smiles information is not necessary (All the existing train examples code does not require it).
So even this is small overhead for computation, I thought it nice to omit when possible.
Also, it uses some amount of memory to keep all the standardized smiles.

chainer_chemistry/dataset/parsers/csv_file_parser.py
```diff
@@ -160,24 +161,12 @@ def parse(self, filepath, retain_smiles=False):
             # Spec not finalized yet for general case
             result = pp.process(filepath)
 
+        smiles_list = numpy.array(smiles_list) if return_smiles else None
```
corochann: I feel weired to name `smiles_list` even when we make it to numpy array.
could you change it to 

```smiles = numpy.array(smiles_list) if return_smiles else None```

or `smiles_array` ?

File: chainer_chemistry/dataset/parsers/sdf_file_parser.py [hunk 4]
```diff
@@ -129,24 +130,12 @@ def parse(self, filepath, retain_smiles=False):
             # Spec not finalized yet for general case
             result = pp.process(filepath)
 
+        smiles_list = numpy.array(smiles_list) if return_smiles else None
```
corochann: same as above.

chainer_chemistry/dataset/parsers/sdf_file_parser.py
mottodora: The objective of this PR is changing parsers's behavior. So it can't be avoided breaking compatibility. Your (2) comment is right. Therefore I propose returning dict or namedtuple that contains dataset and smiles from parser.parse. It makes easy to maintain compatibility.

chainer_chemistry/dataset/parsers/csv_file_parser.py
delta2323: OK, understood.

delta2323:
@mottodora  It is not important but can you remove [WIP] from the title?

chainer_chemistry/dataset/parsers/csv_file_parser.py
mottodora: I'm going to use `smileses`.

delta2323:
> My main concern is default argument. For now, I set True as return_smiles default argument.

I would say it is good to set the default argument `False`, as most of expected use cases do not require SMILES and it reduces memory usage.

mottodora:
OK. I will change it.

chainer_chemistry/dataset/parsers/sdf_file_parser.py
delta2323: OK, understood.

chainer_chemistry/datasets/tox21.py
```diff
@@ -73,11 +73,14 @@ def postprocess_label(label_list):
                            labels=labels)
 
     if return_smiles:
-        train_result = parser.parse(get_tox21_filepath('train'))
+        train_result = parser.parse(get_tox21_filepath('train'),\
+                                    return_smiles=True)
```
corochann: Can you move  `parser.parse(..., return_smiles=return_smiles)` outside of `if return_smiles ... else ... ` clause?
Other `val_result, test_result` as well.

mottodora: OK

PR #99, opened 2018-02-26
Update RSGCN readout

corochann:
1. update readout activation function, 2. support case when readout is type Link.

See discussion at the end of #89

### DIFF
--- a/chainer_chemistry/dataset/preprocessors/rsgcn_preprocessor.py
+++ b/chainer_chemistry/dataset/preprocessors/rsgcn_preprocessor.py
@@ -47,8 +47,10 @@ class RSGCNPreprocessor(MolPreprocessor):
 
         # adjust adjacent matrix
         degree_vec = numpy.sum(adj_array, axis=1)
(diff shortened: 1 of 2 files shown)


corochann:
@anaruse 
Just for your information. 

1. I found bug in your preprocessing
https://github.com/pfnet-research/chainer-chemistry/pull/99/files#diff-d60ebd9ba9c7b3bfa3f127c763acbbc1R50
I think we need to take "inverse" of D, (namely D^{-1/2} in Eq (8) of paper https://arxiv.org/pdf/1609.02907.pdf .) but I guess you forgot to take inverse.

2. I updated readout function as written in the end of PR #89 
But after I updated the bug in 1., I get a result that activation does not affect to performance so much.
So I'm going to set `activation=None` as default behavior.

anaruse:
Thank you for letting me know that.

Re: 1., it was my mistake. I forgot to take inverse ...
Will update https://github.com/pfnet-research/chainer-chemistry/pull/90 based on the changes above.

PR #103, opened 2018-02-26
Schnet predict

corochann:
Support `tox21/inference.py` for SchNet

### DIFF
--- a/examples/tox21/predictor.py
+++ b/examples/tox21/predictor.py
@@ -64,12 +64,11 @@ class GraphConvPredictor(chainer.Chain):
         """
 
         super(GraphConvPredictor, self).__init__()
-        self.mlp = None
         with self.init_scope():
             self.graph_conv = graph_conv
             if isinstance(mlp, chainer.Link):
                 self.mlp = mlp
-        if self.mlp is None:
+        if not isinstance(mlp, chainer.Link):
             self.mlp = mlp
 
     def __call__(self, atoms, adjs):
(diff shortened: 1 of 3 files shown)


examples/tox21/predictor.py
```diff
@@ -78,8 +86,9 @@ def predict(self, atoms, adjs):
 class InferenceLoop(object):
     """Wrapper for predictors that offers inference loop."""
 
-    def __init__(self, predictor):
+    def __init__(self, predictor, batchsize=128):
         self.predictor = predictor
```
mottodora: I feel it is kind to check whether predictor has 'predict' method.
Like `callable(getattr(predictor_, 'predict', None))`
How do you think?

examples/tox21/predictor.py
```diff
@@ -50,23 +52,29 @@ def build_predictor(method, n_unit, conv_layers, class_num):
...
-            mlp: multi layer perceptron, used as final connected layer
+            mlp: multi layer perceptron, used as final connected layer.
+                It can be `None` if no operation is necessary after
+                `graph_conv` calculation.
         """
 
         super(GraphConvPredictor, self).__init__()
         with self.init_scope():
             self.graph_conv = graph_conv
+            if isinstance(mlp, chainer.Link):
```
mottodora: Is this check necessary?

mottodora: Oh sorry. I would like to confirm. If mlp == `None`,  will it raise Error in `with init_scope` statement?

examples/tox21/predictor.py
corochann: I think so.

examples/tox21/predictor.py
corochann: If mlp is not Link (I'm assuming it can be None or even function), then I thought it is better to set this value outside of `init_scope` but I don't know.

PR #104, opened 2018-02-26
Rsgcn qm9

corochann:
Please merge this after PR #103 

1. support RSGCN in QM9 example
2. modified GraphConvPredictor to work on SchNet as well 
3. update argparse to use '-' instead of '_' which is same with tox21 example
4. Change default batchsize=32


mottodora:
LGTM!
I can execute train_qm9.py with RSGCN successfully.

PR #110, opened 2018-03-01
Doc concat mols

corochann:


mottodora:
LGTM!
I run test in GPU environment. It finish successfully.

PR #123, opened 2018-03-29
Print warning message when import error occured at chainer_chemistry.…

ir5:
…datasets to handle a situation when RDKit is not installed

This pull request is intended to solve issue #119.
If RDKit is installed, everything goes as usual.
If not installed,
- A warning message will be shown when we do "import chainer_chemistry". A module chainer_chemistry.datasets will simply be ignored.
- Meanwhile, an error will invoke when we explicitly import chainer_chemistry.datasets.

### DIFF
--- a/chainer_chemistry/__init__.py
+++ b/chainer_chemistry/__init__.py
@@ -1,5 +1,15 @@
+import warnings
+
 from chainer_chemistry import dataset  # NOQA
-from chainer_chemistry import datasets  # NOQA
+try:
+    from chainer_chemistry import datasets  # NOQA
+except ImportError:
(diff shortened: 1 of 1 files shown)


corochann:
LGTM.
Could you review @mottodora ?

mottodora:
Sorry for late response. I found `chainer-chemistry.model` also imports RDKit.
https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/__init__.py#L5

ir5:
Sure, I will fix this PR for `chainer-chemistry.model` as well.

PR #127, opened 2018-04-04
copy weavenet parameter to config.py

### LINKED ISSUE #119: import without install RDKit

mottodora:
resolve #119

### DIFF
--- a/chainer_chemistry/config.py
+++ b/chainer_chemistry/config.py
@@ -4,3 +4,6 @@
 # --- Constant definitions ---
 # The maximum atomic number in rdkit
 MAX_ATOMIC_NUM = 117
+
+# For weave
+DEFAULT_NUM_MAX_ATOMS = 20  # 60  # paper
--- a/chainer_chemistry/config.py
+++ b/chainer_chemistry/config.py
@@ -4,6 +4,4 @@
 # --- Constant definitions ---
 # The maximum atomic number in rdkit
 MAX_ATOMIC_NUM = 117
-
-# For weave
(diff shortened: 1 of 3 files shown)


mottodora:
There are some global variables in weavenet preprocessor. 
For resolving #119(https://github.com/pfnet-research/chainer-chemistry/issues/119), we need only `DEFAULT_NUM_MAX_ATOMS` and I copy it to config.py. (I choose copy for backward compatibility)
If you think that we need to copy all global variables to config.py. I will do that.

delta2323:
It seems odd to me why it resolves #119 because this variable is seemingly irrelevant to RDKit. Could you explain more?

mottodora:
https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/dataset/preprocessors/weavenet_preprocessor.py
There are `import` statements related to `RDKit` in `weavenet_preprocessor`.
Does it make sense to you?

corochann:
Even though we should come up better way to handle these dependency, as a workaround (maybe for temporary), I think it is better to merge it at first.
How about change the name to `WEAVE_DEFAULT_NUM_MAX_ATOMS ` and refer from both model and preprocessor?

delta2323:
The proposal from @corochann is acceptable to me. Once I thought adding constants specific to some algorithm seem unnatural to me. But we do not have appropriate place to put such constants other than `config`, as an algorithm consists of several components (preprocessor, Chain instance etc.). Therefore, adding constants to `config` might be a good temporary answer.

mottodora:
I fix now.
Can you review?

PR #128, opened 2018-04-06
[WIP] Init serializable chain

corochann:


corochann:
We have discussed internally that just using `pickle` is fine in our use case.
So we introduce pickle save/load support.
https://github.com/pfnet-research/chainer-chemistry/pull/139

This PR may be simply closed.

PR #131, opened 2018-04-10
Parse index

corochann:
Add `target_index` option to enable partial parse

### DIFF
--- a/chainer_chemistry/dataset/parsers/csv_file_parser.py
+++ b/chainer_chemistry/dataset/parsers/csv_file_parser.py
@@ -13,6 +13,18 @@ from chainer_chemistry.datasets.numpy_tuple_dataset import NumpyTupleDataset
 import traceback
 
 
+def _read_csv(filepath):
+    try:
+        # It is recommended to use `read_csv` method in pandas version
+        # after 0.18.x
+        df = pandas.read_csv(filepath)
+    except AttributeError as e:
+        # It is deprecated in newer versions of pandas, but we use
+        # this method for older version of pandas.
+        df = pandas.DataFrame.from_csv(filepath)
+    return df
+
+
 class CSVFileParser(BaseFileParser):
     """csv file parser
 
@@ -69,14 +81,7 @@ class CSVFileParser(BaseFileParser):
 
         # counter = 0
         if isinstance(pp, MolPreprocessor):
-            try:
-                # It is recommended to use `read_csv` method in pandas version
-                # after 0.18.x
-                df = pandas.read_csv(filepath)
-            except AttributeError as e:
-                # It is deprecated in newer versions of pandas, but we use
-                # this method for older version of pandas.
-                df = pandas.DataFrame.from_csv(filepath)
+            df = _read_csv(filepath)
 
             if target_index is not None:
                 df = df.iloc[target_index]
@@ -177,3 +182,7 @@ class CSVFileParser(BaseFileParser):
             if self.postprocess_fn is not None:
                 result = self.postprocess_fn(result)
             return {"dataset": NumpyTupleDataset(result), "smiles": smileses}
+
+    def extract_total_num(self, filepath):
+        df = _read_csv(filepath)
+        return len(df)
--- a/chainer_chemistry/dataset/parsers/csv_file_parser.py
+++ b/chainer_chemistry/dataset/parsers/csv_file_parser.py
@@ -13,18 +13,6 @@ from chainer_chemistry.datasets.numpy_tuple_dataset import NumpyTupleDataset
 import traceback
 
 
-def _read_csv(filepath):
-    try:
-        # It is recommended to use `read_csv` method in pandas version
-        # after 0.18.x
-        df = pandas.read_csv(filepath)
-    except AttributeError as e:
-        # It is deprecated in newer versions of pandas, but we use
-        # this method for older version of pandas.
-        df = pandas.DataFrame.from_csv(filepath)
-    return df
-
-
 class CSVFileParser(BaseFileParser):
     """csv file parser
 
@@ -81,7 +69,7 @@ class CSVFileParser(BaseFileParser):
 
         # counter = 0
         if isinstance(pp, MolPreprocessor):
-            df = _read_csv(filepath)
+            df = pandas.read_csv(filepath)
 
             if target_index is not None:
                 df = df.iloc[target_index]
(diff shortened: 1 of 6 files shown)


chainer_chemistry/dataset/parsers/csv_file_parser.py
```diff
@@ -13,6 +13,18 @@
...
+def _read_csv(filepath):
+    try:
+        # It is recommended to use `read_csv` method in pandas version
+        # after 0.18.x
+        df = pandas.read_csv(filepath)
+    except AttributeError as e:
+        # It is deprecated in newer versions of pandas, but we use
+        # this method for older version of pandas.
+        df = pandas.DataFrame.from_csv(filepath)
+    return df
```
delta2323: It is not directly related to this PR but should we write a policy somewhere that we do not test older versions of dependent modules (like Pandas) except Python itself, although we sometimes do make code work with them, as we do here?

chainer_chemistry/dataset/parsers/csv_file_parser.py
```diff
@@ -13,6 +13,18 @@
 import traceback
 
 
+def _read_csv(filepath):
+    try:
+        # It is recommended to use `read_csv` method in pandas version
+        # after 0.18.x
+        df = pandas.read_csv(filepath)
```
delta2323: It seems Pandas v0.15.0 already has `read_csv` (See [release note](https://github.com/pandas-dev/pandas/releases/tag/v0.15.0)). I thought it is safe to assume Pandas in most environment has this method, as v0.15.2. But let me know if you know some common environment in which this assumption does not hold.
 

```
In [1]: import pandas as pd
In [2]: pd.read_csv
Out[2]: <function pandas.io.parsers.read_csv>
In [3]: pd.__version__
Out[3]: '0.15.2'
```

chainer_chemistry/dataset/parsers/csv_file_parser.py
```diff
@@ -172,3 +182,7 @@ def parse(self, filepath, return_smiles=False):
             if self.postprocess_fn is not None:
                 result = self.postprocess_fn(result)
             return {"dataset": NumpyTupleDataset(result), "smiles": smileses}
+
+    def extract_total_num(self, filepath):
+        df = _read_csv(filepath)
+        return len(df)
```
delta2323: When `parse` method failed feature extraction from some rows in a CSV file, the final dataset size is different from the value this method returns. Whether it is acceptable or not depends on how we use this method, I would like to know your expected use case.

chainer_chemistry/dataset/parsers/sdf_file_parser.py
```diff
@@ -53,12 +55,19 @@ def parse(self, filepath, return_smiles=False):
...
+            if target_index is None:
+                target_index = numpy.arange(len(mol_supplier))
+
             features = None
 
             total_count = len(mol_supplier)
             fail_count = 0
             success_count = 0
-            for mol in tqdm(mol_supplier):
+            # for mol in tqdm(mol_supplier):
```
delta2323: Remove unused commented code.

chainer_chemistry/dataset/parsers/sdf_file_parser.py
```diff
@@ -53,12 +55,19 @@ def parse(self, filepath, return_smiles=False):
         if isinstance(pp, MolPreprocessor):
             mol_supplier = Chem.SDMolSupplier(filepath)
 
+            if target_index is None:
+                target_index = numpy.arange(len(mol_supplier))
```
delta2323: If `mol_supplier` does not accept `numpy.integer`, I think we do not have to prepare `target_index` as a `numpy.ndarray` here.

chainer_chemistry/dataset/parsers/csv_file_parser.py
delta2323: I asked @corochann directly and was told that he expect we use this method to determine the value fed to `target_index` option of `CSVFileParser.parse`. For example, if we want to extract input feature from 10% of whole dataset, we need to know how many samples are in a file. Therefore, the returned value of `extract_total_num` need not to be same as the final dataset size.

chainer_chemistry/dataset/parsers/sdf_file_parser.py
corochann: I changed this place.

But I'd like to keep 
```
# `mol_supplier` does not accept numpy.integer, we must use int
mol = mol_supplier[int(index)]
```

corochann:
updated based on comment. could you review again?

chainer_chemistry/dataset/parsers/csv_file_parser.py
delta2323: I imagined to make a consensus within a development team and write it somewhere in the [development policy](http://chainer-chemistry.readthedocs.io/en/stable/development.html).

chainer_chemistry/dataset/parsers/csv_file_parser.py
```diff
@@ -184,5 +172,19 @@ def parse(self, filepath, return_smiles=False, target_index=None):
             return {"dataset": NumpyTupleDataset(result), "smiles": smileses}
 
     def extract_total_num(self, filepath):
-        df = _read_csv(filepath)
+        """Extracts total number of data which can be parsed
+
+        This method is to determine the value fed to `target_index` option of
+        `parse` method. For example, if we want to extract input feature from
+        10% of whole dataset, we need to know how many samples are in a file.
+        The returned value of thid method need not to be same as the final
```
delta2323: I think "may not" is more preferable than "need not".

delta2323: typo: thid -> this

chainer_chemistry/dataset/parsers/csv_file_parser.py
```diff
@@ -184,5 +172,19 @@ def parse(self, filepath, return_smiles=False, target_index=None):
             return {"dataset": NumpyTupleDataset(result), "smiles": smileses}
 
     def extract_total_num(self, filepath):
-        df = _read_csv(filepath)
+        """Extracts total number of data which can be parsed
+
+        This method is to determine the value fed to `target_index` option of
```
delta2323: I would suggest to rephrase as "We can use this method to determine the value fed to ..." because the usage described here is not the only way to use this method (user may use this method in different way).

chainer_chemistry/dataset/parsers/sdf_file_parser.py
```diff
@@ -154,5 +153,19 @@ def parse(self, filepath, return_smiles=False, target_index=None):
             return {"dataset": NumpyTupleDataset(result), "smiles": smileses}
 
     def extract_total_num(self, filepath):
+        """Extracts total number of data which can be parsed
+
+        This method is to determine the value fed to `target_index` option of
+        `parse` method. For example, if we want to extract input feature from
+        10% of whole dataset, we need to know how many samples are in a file.
+        The returned value of thid method need not to be same as the final
+        dataset size.
```
delta2323: Please apply same changes to this document as well.

corochann:
reflected comments.

delta2323:
LGTM. I'm running the example tests.

corochann:
Please merge both PRs after you checked it ok.

Issue #134, opened 2018-04-12
Improve classifier


mottodora:
- [x] Add document about what is different from Classifier in Chainer to the document.
https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/models/prediction/classifier.py#L34-L35
- [x] The place of L23 is strange.
https://github.com/pfnet-research/chainer-chemistry/blob/master/examples/tox21/predict_tox21.py#L23
- [x] There are `predict_tox21.py` and  `inference_tox21.py`. They are very similary name.

PR #136, opened 2018-04-13
Predict example refactor

corochann:
For the last item of issue: https://github.com/pfnet-research/chainer-chemistry/issues/134


mottodora:
LGTM!

PR #139, opened 2018-04-16
Pickle BaseForwardModel

corochann:
implement save_pickle, load_pickle and added test

### DIFF
--- a/chainer_chemistry/models/prediction/base.py
+++ b/chainer_chemistry/models/prediction/base.py
@@ -1,3 +1,5 @@
+import pickle
+
 import chainer
 from chainer.dataset.convert import concat_examples
 from chainer import link, cuda
@@ -19,9 +21,9 @@ def _extract_numpy(x):
 
 class BaseForwardModel(link.Chain):
 
-    """A base model which supports _forward functionality.
+    """A base model which supports forward functionality.
 
-    It also supports `device` id management.
+    It also supports `device` id management and pickle save/load functionality.
 
     Args:
         device (int): GPU device id of this model to be used.
@@ -127,3 +129,43 @@ class BaseForwardModel(link.Chain):
             return result[0]
         else:
             return result
+
+    def save_pickle(self, filepath, protocol=None):
+        """Save the model to `filepath` as a pickle file
+
+        Args:
+            filepath (str): file path of pickle file.
+            protocol (int or None): protocol version used in `pickle`.
+                Use 2 if you need python2/python3 compatibility.
+                3 or higher is used for python3.
+
+        """
+        current_device = self.get_device()
+
+        # --- Move the model to CPU for saving ---
+        self.update_device(-1)
+        with open(filepath, mode='wb') as f:
+            pickle.dump(self, f, protocol=protocol)
+
+        # --- Revert the model to original device ---
+        self.update_device(current_device)
+
+    @staticmethod
+    def load_pickle(filepath, device=-1):
+        """Load the model from `filepath` of pickle file, and send to `device`
+
+        Args:
+            filepath (str): file path of pickle file.
+            device (int): GPU device id of this model to be used.
+                -1 indicates to use in CPU.
+
+        """
+        with open(filepath, mode='rb') as f:
+            model = pickle.load(f)
+
+        if not isinstance(model, BaseForwardModel):
+            raise TypeError('Unexpected type {}'.format(type(model)))
+
+        # --- Revert the model to specified device ---
+        model.initialize(device)
+        return model
--- a/tests/models_tests/prediction_tests/test_base.py
+++ b/tests/models_tests/prediction_tests/test_base.py
@@ -1,0 +1,59 @@
+import os
+
+import chainer
+from chainer import cuda
+import numpy
+import pytest
+import tempfile
+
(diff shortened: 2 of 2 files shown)


delta2323:
Could you check the result of Travis?

delta2323:
Please write unit tests for failed case, that is, `DummyForwardModel` correctly raises errors when a picklable object is serialized with `save` and a corrupted file is unpickled with `load`.

chainer_chemistry/models/prediction/base.py
```diff
@@ -127,3 +129,43 @@ def _forward(self, data, fn, batchsize=16,
...
             return result
+
+    def save_pickle(self, filepath, protocol=None):
+        """Save the model to `filepath` as a pickle file
+
+        Args:
+            filepath (str): file path of pickle file.
+            protocol (int or None): protocol version used in `pickle`.
+                Use 2 if you need python2/python3 compatibility.
+                3 or higher is used for python3.
```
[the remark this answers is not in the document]
corochann: That is why i am writing "3 or higher". But i can add link as well if necessary.

corochann:
> DummyForwardModel correctly raises errors when a picklable object is serialized with save and a corrupted file is unpickled with load.

What do we want to support? 
I would say the error handling when "corrupted file is unpickled with load" is pickle's work, and not our library's work, what do you think?

corochann:
Writing a test means we will define a spec to show some error message in the case of load.
But I think it is ok to let `pickle` to handle these case, not our library's scope.

corochann:
Updated.

chainer_chemistry/models/prediction/base.py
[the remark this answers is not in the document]
delta2323: I see. I misunderstood the sentence.

delta2323:
> I would say the error handling when "corrupted file is unpickled with load" is pickle's work, and not our library's work, what do you think?

I know we do not have to test `pickle` itself, assuming  that `pickle` works correctly. I wanted to specify how to deal with errors raised by `pickle` in the save/load methods. For example, currently we leave the `pickle` errors as they are. But one of alternative choices we can take is to grasp the pickle error.

mottodora:
LGTM.
However, I feel we need to improve documentation.
1.  `load_pickle` function is a bit tricky. It is kind that there is a example in document.
2. We need to stress that `load_pickle` sometimes fails because of version issue and more.
Additionally, I can understand @delta2323's suggestion. But all of them are not critical. So I would like to merge first and I'm going to make issues and solve them by other PRs.

delta2323:
Comments on @mottodora 's latest suggestion:

1. I'm for careful documenting. Especially, I think we should note that we assume users save and load models with a same version and that we do not guarantee model compatibility across different ones, that is, a pickled model saved in an older version may not be loaded newer one and vice versa. If we adopt this policy, I think it would be better to add it to the [compatibility policy]. Also, it would be better to add version number to saved model. If we would apply this policy strictly, we can forbid load models that have different version number. But it seems too much to me.

One concern is that this restriction might motivate users to remain older versions, which may be the opposite direction to our aggressive version-increment policy. 

2. It is acceptable to write the test afterwards.

corochann:
Added documents. Please merge if it is ok.

chainer_chemistry/models/prediction/base.py
```diff
@@ -134,6 +134,38 @@ def _forward(self, data, fn, batchsize=16,
     def save_pickle(self, filepath, protocol=None):
         """Save the model to `filepath` as a pickle file
 
+        This function send the parameters to CPU before saving the model so
+        that the pickled file can be loaded with in CPU-only environment.
```
delta2323: I think it may remind users that this method has a side effect and parameters are in CPU after the method is called. But actually, it moves back to an original device. So, I would suggest to describe it.

chainer_chemistry/models/prediction/base.py
```diff
@@ -134,6 +134,38 @@ def _forward(self, data, fn, batchsize=16,
...
 
+        This function send the parameters to CPU before saving the model so
+        that the pickled file can be loaded with in CPU-only environment. 
+
+        Saved pickle file can be loaded with `load_pickle` static method.
+
+        Note that the transportability of the saved file follows the
+        specification of `pickle` module, namely serialized data depends on the
+        specific class or attribute structure when saved. The file may not be
+        loaded in different environment (python version is different or
```
delta2323: I thinks description in the parenthesis is a bit redundant. How about `(versions of Python or dependent libraries)`?

chainer_chemistry/models/prediction/base.py
```diff
@@ -157,6 +189,18 @@ def save_pickle(self, filepath, protocol=None):
...
+        The file saved by `save_pickle` method can be loaded, but it may fail
+        to load when loading from different develop environment or after
+        updating library version.
+        See `save_pickle` method for the transportability of the saved file.
+
+    .. admonition:: Example
+
+       >>> from chainer_chemistry.models import BaseForwardModel
+       >>> filepath = 'model.pkl'
+       >>> # `load_pickle` is static method, call from Class to get instance.
```
delta2323: get *an* instance.

delta2323:
LGTM except the comments.

Issue #142, opened 2018-04-18
GGNN output is not graph isomorphism invariant


corochann:
It is reported that GGNN output differs when we permute the index of atoms/adjs... 
We should also check other network behavior as well.

PR #143, opened 2018-04-18
Ggnn graph invariant

corochann:
resolves https://github.com/pfnet-research/chainer-chemistry/issues/142


chainer_chemistry/models/ggnn.py
```diff
@@ -66,9 +66,9 @@ def update(self, h, adj, step=0):
         message_layer_index = 0 if self.weight_tying else step
         m = functions.reshape(self.message_layers[message_layer_index](h),
                               (mb, atom, out_ch, self.NUM_EDGE_TYPE))
-        # m: (minibatch, ch, atom, edge_type)
+        # m: (minibatch, atom, ch, edge_type)
         # Transpose
-        m = functions.transpose(m, (0, 3, 2, 1))
+        m = functions.transpose(m, (0, 3, 1, 2))
```
corochann: This is the only mis-implementation. (Other implementation is only for test)

PR #144, opened 2018-04-20
Run examples using Python 2.7

mihaimorariu:


chainer_chemistry/datasets/numpy_tuple_dataset.py
```diff
@@ -89,7 +89,7 @@ def save(cls, filepath, numpy_tuple_dataset):
     def load(cls, filepath):
         if not os.path.exists(filepath):
             return None
-        load_data = numpy.load(filepath)
+        load_data = numpy.load(filepath, encoding='latin1')
```
corochann: Is this really necessary for python2?

chainer_chemistry/datasets/tox21.py
```diff
@@ -107,20 +107,25 @@ def postprocess_label(label_list):
...
     Args:
         dataset_type(str): Name of the target dataset type.
             Either 'train', 'val', or 'test'.
 
-    Returns (str): filepath for the tox21 dataset
+    Returns (str): file path for the tox21 dataset
 
     """
+    valid_dataset_types = {'train', 'val', 'test'}
+    if not dataset_type in valid_dataset_types:
```
corochann: Could you use already defined constant value `_config.keys()` instead of defining again here?

chainer_chemistry/datasets/tox21.py
```diff
@@ -107,20 +107,25 @@ def postprocess_label(label_list):
...
+        raise ValueError('Invalid dataset type \'{}\'. Accepted values are '
+            '\'train\', \'val\' or \'test\'.'.format(dataset_type))
```
corochann: Personally, I think we do not need to round with \\'

Other way is to use "string here with 'string' hoge" <- this kind of format.

corochann: Also, it is not mandatory but if possible please check pep8.
It accepts line length less than 80 chars.
http://chainer-chemistry.readthedocs.io/en/latest/development.html#coding-guideline

chainer_chemistry/datasets/tox21.py
```diff
@@ -107,20 +107,24 @@ def postprocess_label(label_list):
...
         dataset_type(str): Name of the target dataset type.
             Either 'train', 'val', or 'test'.
 
-    Returns (str): filepath for the tox21 dataset
+    Returns (str): file path for the tox21 dataset
 
     """
+    if dataset_type not in _config.keys():
+        raise ValueError("Invalid dataset type '{}'. Accepted values are "
+                         "'train', 'val' or 'test'.".format(dataset_type))
```
corochann: Adding this check means that we will add a spec to raise appropriate error when dataset_type is wrong.
To check this, I appreciate if you can implement test to catch this valueerror when wrong dataset_type is passed.

Test is implemented at here.
https://github.com/pfnet-research/chainer-chemistry/blob/master/tests/datasets_tests/test_tox21.py

implementation will be like this
https://github.com/pfnet-research/chainer-chemistry/blob/81b18c6d1b281dbe66673b7d9a8117aff417af77/tests/dataset_tests/preprocessors_tests/test_ggnn_preprocessor.py#L32-L33

corochann:
LGTM except the comment, can you write test? If you are busy, we can skip it and I will merge it at first.

PR #147, opened 2018-04-21
Model evaluation on tox21

n-yoshikawa:
This pull request adds an evaluation of each model in tox21.

I want to discuss:
- Where to put graph images (in this repository or other place)
- How to show the log of experiment


n-yoshikawa:
I updated the code, please review.

examples/tox21/evaluate_models_tox21.sh
```diff
@@ -0,0 +1,17 @@
...
+gpu=-1
+methods=(nfp ggnn schnet weavenet rsgcn)
+results=()
+
+mkdir -p results-evaluation
+for method in ${methods[@]}
+do
+    python train_tox21.py --method ${method} --gpu ${gpu}
+    python predict_tox21_with_classifier.py > log-${method}
+    roc_auc=`cat log-${method} | tail -1 | grep -oP '0\.\d+'`
```
corochann: I have a concern in this line that it relies on the fact that last line of the `predict_tox21_with_classifier.py` is the evaluation result, which may not be guaranteed after the code `predict_tox21_with_classifier.py` refactored.

corochann: How about modifying `predict_tox21_with_classifier.py` to save the evaluation result in some text file (json file etc), and load this text file in `plot.py`?

corochann:
@n-yoshikawa @mottodora @delta2323

Is it ok to merge this PR? I guess it will show the png file for the evaluation result in the top page of README.  

We should decide...
> Where to put graph images (in this repository or other place)
> How to show the log of experiment

also, I want to put in table that 

* Is it ok to show only validation result? How about showing both 'train' & 'validation' result to see which model more likely overfit or etc.
* Currently hyper parameter is not tuned or not following author's implementation. several network may improve easily by modify some parameters. how to handle this?

delta2323:
Should we run `evaluate_models_tox21.sh` in the test scripts (`example_test_*pu.sh`)?

delta2323:
I'm running `evaluate_models_tox21.sh` in my environment.

delta2323:
Please add dependency of `plot.py` on `seaborn` in `README`.

README.md
```diff
@@ -108,6 +108,11 @@ Currently, following dataset is supported.
 - QM9 [7, 8]
 - Tox21 [9]
 
+## Evaluation
+![tox21-evaluation](https://user-images.githubusercontent.com/29328746/39457687-974ad4ce-4d2a-11e8-9305-f6ca607af9dc.png)
+
+This image is the result of `examples/tox21/evaluate_models_tox21.sh`.
```
delta2323: I think we should note that we do not guarantee reproducibility of papers in which each model proposed and hence result doesn't imply performance of models themselves. For example, SchNet performs worst in the experiment, but it might be faults in our implementation.

corochann: Yes I agree.

delta2323: @n-yoshikawa If you agree, could you add the aforementioned comments?

n-yoshikawa: I added the comment.

delta2323:
Basically OK, could you resolve confclicts?

delta2323:
LGTM

PR #149, opened 2018-04-22
bug fix

corochann:


delta2323:
Could you let us know about the bug you want to eliminate?

corochann:
1. retain_smiles -> return_smiles in get_tox21()
otherwise, code does not work at all.

2. train_tox21 works only for 'rsgcn' and 'sparse_rsgcn' method for this branch.
Since, `multiple` attributed is added as hacking way in this branch.

corochann:
Also, I'd like to ignore travis check of this branch.

delta2323:
> I would like to consider integrating this implementation to master branch after chainer is merged.

I agree.

corochann:
> Also, I'd like to ignore travis check of this branch.

I meant just allow this PR to be merged even the Travis fails?

PR #154, opened 2018-04-23
Remove Chainer v2 test from Travis

delta2323:
Since Chainer v4 has been released, Chainer installed by `CHAINER_VERSION="chainer` in travis has been changed to from v4 to v3. That means Chainer v3 is not tested in the current configuration. This PR adds a test with Chainer v4 to `.travis.yml`.

I did not remove configuration about Chainer v2 (`CHAINER_VERSION="chainer<3`)  because testing time in Travis is not a bottleneck in the current workflow. But I would be like opinions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ python:
   - 3.6
 env:
   - CHAINER_VERSION="chainer<3"
+  - CHAINER_VERSION="chainer<4"
   - CHAINER_VERSION="chainer"
   - CHAINER_VERSION="prerelease"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ os: linux
 dist: trusty
 python:
   - 2.7
-  - 3.5
   - 3.6
 env:
   - CHAINER_VERSION="chainer<3"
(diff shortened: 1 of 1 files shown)


corochann:
I think building `miniconda` takes most of the time. Does it build `miniconda` 12 times?

delta2323:
> Does it build miniconda 12 times?

Yes, in total. The reason for using miniconda is that it is recommended way to install RDKit via it (if I remember correctly) and we expect most users install in this way.

delta2323:
I feel sometimes many jobs are queued (e.g. before releasing a new version), so it is better to reduce testing time. 

Currently we test with v2, 4, and 5, I think at least we should replace v2 with v3, if we do not want to increase the number of configurations in the CI.

delta2323:
An alternative way is to remove testing with Python 3.5.

delta2323:
I remove tests with Python 3.5 as to keep testing time. It is just a work around and we need to consider a better way.

delta2323:
After offline discussion, we support the stable version of Chainer (v4 now) and test the stable and develop version of Chainer (v4 and v5) in Travis. So, the resulting change is to disable Chainer v2. I changed the PR title accordingly. Rel. #190.

delta2323:
Let me close this PR and resend it.

PR #157, opened 2018-04-24
supress value error when label contains one label

corochann:


chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -73,6 +74,10 @@ class ROCAUCEvaluator(Evaluator):
             are considered as negative.
         ignore_labels (int or list or None): labels to be ignored.
             `None` is used to not ignore all labels.
+        raise_value_error (bool): If False, `ValueError` caused by
+            `roc_auc_score` calculation is suppressed and ignored with warning
```
delta2323: with *a* warning message.

chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -73,6 +74,10 @@ class ROCAUCEvaluator(Evaluator):
             are considered as negative.
         ignore_labels (int or list or None): labels to be ignored.
             `None` is used to not ignore all labels.
+        raise_value_error (bool): If False, `ValueError` caused by
```
delta2323: enclose `False` with backquotes.

chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -132,7 +140,19 @@ def evaluate(self):
...
-        roc_auc = metrics.roc_auc_score(t_total, y_total)
+        try:
+            roc_auc = metrics.roc_auc_score(t_total, y_total)
+        except ValueError as e:
+            if self.raise_value_error:
+                raise e
+            else:
+                # This is usually caused by the following
+                # Only one class present in y_true.
+                # ROC AUC score is not defined in that case
```
delta2323: It describe when `roc_auc_score` is raised. So I think we should move to l.145.

corochann:
updated

chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -132,7 +140,19 @@ def evaluate(self):
         # --- set positive labels to 1, negative labels to 0 ---
         pos_indices = numpy.in1d(t_total, self.pos_labels)
         t_total = numpy.where(pos_indices, 1, 0)
-        roc_auc = metrics.roc_auc_score(t_total, y_total)
+        try:
+            roc_auc = metrics.roc_auc_score(t_total, y_total)
+        except ValueError as e:
+            # `ValueError` is usually caused by the following
```
delta2323: It seems you list only one situation where `ValueError` is raised. So, `by the following` seems strange.

chainer_chemistry/training/extensions/roc_auc_evaluator.py
```diff
@@ -132,7 +140,19 @@ def evaluate(self):
...
+            # Only one class present in y_true.
```
delta2323: We should not capitalize here because the previous sentence is not completed.

corochann:
ROCAUCEvaluator will raise ValueError when only one label is contained in the dataset.
This will stop the whole script, but this is evaluation script and not affect to training.
I added a option to ignore this error so that we can keep running the training.

It has another benefit that we can make test example faster.

corochann:
Could you review @delta2323 @mottodora ?

delta2323:
LGTM

PR #158, opened 2018-04-24
Test example faster

corochann:
ROCAUCEvaluator will raise ValueError when only one label is contained in the dataset.
This will stop the whole script, but this is evaluation script and not affect to training.
I added a option to ignore this error so that we can keep running the training. 

It has another benefit that we can make test example faster.


corochann:
Please merge after #157

delta2323:
Now #157 is merged. Could you merge or rebase the latest master branch?

corochann:
merged

corochann:
Can you merge if it is ok? @delta2323 @mottodora

delta2323:
LGTM

PR #163, opened 2018-04-27
Add unit test for MLP

mihaimorariu:


tests/models_tests/test_mlp.py
```diff
@@ -0,0 +1,62 @@
...
+@pytest.fixture
+def model():
+    return MLP(out_dim=out_dim)
+
+
+@pytest.fixture
+def data():
+    atom_data = numpy.random.randint(
+        0, high=MAX_ATOMIC_NUM, size=(batch_size, atom_size)
+    ).astype(numpy.float32)
```
corochann: input of MLP is expected to be a hidden feature vector (=random float value), instead of embed vector (int value of atomic number).

corochann:
LGTM except above comment.

PR #171, opened 2018-05-14
use regressor class

corochann:


examples/own_dataset/train.py
```diff
@@ -2,24 +2,27 @@
...
+import os
 import sys
 
 import chainer
 from chainer.datasets import split_dataset_random
 from chainer import functions as F, cuda, Variable
-from chainer import iterators as I
+from chainer import iterators
 from chainer import links as L
-from chainer import optimizers as O
```
corochann: flake8 said I and O are ambiguous

examples/own_dataset/train.py
```diff
@@ -156,31 +159,35 @@ def scaled_abs_error(x0, x1):
             diff = cuda.to_cpu(x0) - cuda.to_cpu(x1)
         return numpy.mean(numpy.absolute(diff), axis=0)[0]
 
-    classifier = L.Classifier(model, lossfun=F.mean_squared_error,
-                              accfun=scaled_abs_error)
+    regressor = Regressor(
+        model, lossfun=F.mean_squared_error,
+        metrics_fun={'abs_error': scaled_abs_error},
```
corochann: use Regressor class with specified key for metrics.

examples/own_dataset/train.py
```diff
@@ -156,31 +159,35 @@ def scaled_abs_error(x0, x1):
...
-    trainer.extend(E.PrintReport(['epoch', 'main/loss', 'main/accuracy',
+    # Note that original scale absolute errors are reported in
+    # (validation/)main/abs_error
+    trainer.extend(E.PrintReport(['epoch', 'main/loss', 'main/abs_error',
                                   'validation/main/loss',
-                                  'validation/main/accuracy',
+                                  'validation/main/abs_error',
                                   'elapsed_time']))
     trainer.extend(E.ProgressBar())
     trainer.run()
```
corochann: Added example for how to save model parameters for future usage.

n-yoshikawa:
LGTM

examples/own_dataset/train.py
```diff
@@ -139,48 +166,43 @@ def postprocess_label(label_list):
...
 
+    # --- save regressor's parameters ---
+    protocol = args.protocol
+    save_model_path = os.path.join(args.out, 'model.npz')
+    print('saving trained model to {}'.format(save_model_path))
+    serializers.save_npz(save_model_path, regressor)
+    if args.scale == 'standardize':
+        with open(os.path.join(args.out, 'ss.pkl'), mode='wb') as f:
```
delta2323: Could you reconsider its name so that we can grasp its meaning more easily?

examples/own_dataset/README.md
```diff
@@ -1,4 +1,10 @@
 # Example of using your own dataset
+
+This example shows how to train your own dataset stored in CSV format.
```
delta2323: ... how to train models *with* your own dataset ...

examples/own_dataset/README.md
```diff
@@ -1,4 +1,10 @@
...
+
+This example is for regression task using `Regressor`.
```
delta2323: I would suggest:

We will solve a regression task with `Regressor`. For a classification setting that makes use of `Classifier`, please refer to the `tox21` example.

delta2323: Please add links to the document of `Classifier` and `Regressor`.

http://chainer-chemistry.readthedocs.io/en/stable/generated/chainer_chemistry.models.Classifier.html#chainer_chemistry.models.Classifier
http://chainer-chemistry.readthedocs.io/en/stable/generated/chainer_chemistry.models.Regressor.html#chainer_chemistry.models.Regressor

examples/own_dataset/train.py
```diff
@@ -50,6 +54,26 @@ def __call__(self, atoms, adjs):
         return x
 
 
+class ScaledAbsError(object):
+
+    def __init__(self, scale='standardize', ss=None):
+        self.scale = scale
```
delta2323: Please raise `ValueError` when `scale` is neither `standardize` nor `none`.

examples/own_dataset/train.py
```diff
@@ -50,6 +54,26 @@ def __call__(self, atoms, adjs):
...
+        self.ss = ss
```
delta2323: Could you reconsider the attribute name so that users can understand its functionality from its name?

examples/own_dataset/train.py
```diff
@@ -50,6 +54,26 @@ def __call__(self, atoms, adjs):
...
+
+    def __call__(self, x0, x1):
+        if isinstance(x0, Variable):
+            x0 = cuda.to_cpu(x0.data)
+        if isinstance(x1, Variable):
+            x1 = cuda.to_cpu(x1.data)
+        if self.scale == 'standardize':
+            scaled_x0 = self.ss.inverse_transform(cuda.to_cpu(x0))
```
delta2323: It raises a error when `self.ss` is `None`, which happens when the `ss` option is a default value.

examples/own_dataset/train.py
```diff
@@ -139,48 +166,43 @@ def postprocess_label(label_list):
...
                                   'validation/main/loss',
-                                  'validation/main/accuracy',
+                                  'validation/main/abs_error',
                                   'elapsed_time']))
     trainer.extend(E.ProgressBar())
     trainer.run()
 
+    # --- save regressor's parameters ---
+    protocol = args.protocol
+    save_model_path = os.path.join(args.out, 'model.npz')
```
delta2323: It would be my taste but `save_model_path` looks to me that it indicates an act of saving a path to a model. How about `model_path`?

examples/own_dataset/train.py
```diff
@@ -139,48 +166,43 @@ def postprocess_label(label_list):
...
+    print('saving trained model to {}'.format(save_model_path))
+    serializers.save_npz(save_model_path, regressor)
+    if args.scale == 'standardize':
+        with open(os.path.join(args.out, 'ss.pkl'), mode='wb') as f:
+            pickle.dump(ss, f, protocol=protocol)
```
delta2323: It seems more natural to me that we implement `serialize` method to `Regressor` which serialize its `metrics_fun` attribute (if functions have `serialize` method), instead of saving `ss` apart from `regressor`. How do you think?

corochann: That would be nice!
Can you implement to extend `Regressor`/`Classifier` class to serialize its `metrics_fun` attributes, and serializable `StandardScaler`?
I think it is nice if `Scaler` class (something used in preprocessing/postprocessing) can be serialized.
I think this will be another PR.

corochann: I am not sure `Scaler` class should be an instance of `Link` or not.

delta2323: If I understand correctly, `Serializer` and `Deserializer` can serialize any object that implements `serialize` method.

delta2323: For example, `SerialIterator` has `serialize` method to perpetuate its status (cf. [here](https://github.com/chainer/chainer/blob/master/chainer/iterators/serial_iterator.py#L122)).

examples/own_dataset/train.py
corochann: done

examples/own_dataset/README.md
corochann: done

examples/own_dataset/README.md
corochann: done

examples/own_dataset/train.py
corochann: I removed scale option, only check scaler

examples/own_dataset/train.py
corochann: changed to scaler

examples/own_dataset/train.py
corochann: I removed scale option, only check scaler

examples/own_dataset/train.py
corochann: done

examples/own_dataset/train.py
corochann: todo future

corochann:
Updated. @delta2323 @n-yoshikawa 
I also set default values to `--datafile` and `label` argument so that user do not need to specify it to call
`python train.py` to test in default use case.

examples/own_dataset/train.py
```diff
@@ -50,17 +53,37 @@ def __call__(self, atoms, adjs):
...
+
 def main():
     # Supported preprocessing/network list
     method_list = ['nfp', 'ggnn', 'schnet', 'weavenet', 'rsgcn']
     scale_list = ['standardize', 'none']
 
     parser = argparse.ArgumentParser(
         description='Regression with own dataset.')
-    parser.add_argument('datafile', type=str)
+    parser.add_argument('--datafile', type=str, default='dataset.csv')
```
delta2323: This breaks the example test scripts. Because `datafile` is not mandatory, a superfluous argument is given. Here is an error message:

```
%(pr-171)[U]bash -x example_test_cpu.sh 
...
+ python train.py dataset.csv --method nfp --label value1 --conv-layers 1 --gpu -1 --epoch 1 --unit-num 10 --batchsize 32
usage: train.py [-h] [--datafile DATAFILE]
                [--method {nfp,ggnn,schnet,weavenet,rsgcn}]
                [--label LABEL [LABEL ...]] [--scale {standardize,none}]
                [--conv-layers CONV_LAYERS] [--batchsize BATCHSIZE]
                [--gpu GPU] [--out OUT] [--epoch EPOCH] [--unit-num UNIT_NUM]
                [--seed SEED] [--train-data-ratio TRAIN_DATA_RATIO]
                [--protocol PROTOCOL]
train.py: error: unrecognized arguments: dataset.csv
```
IMHO, we do not have to set the default value as long as usage is written in `README.md` at the cost of breaking existing user code which invokes this script. How do you think?

corochann:
updated

delta2323:
LGTM

PR #183, opened 2018-06-08
Make the 'plot.py' script dataset-independent.

mihaimorariu:
Since the AUC evaluation is going to be applied to multiple datasets, the plotting of ROC-AUC should not be dataset-specific.


corochann:
I think either of below is better.

1. keep current design (we need to re-define similar function in each example)
2. make util function inside library to reuse the visualization

delta2323:
@corochann Can I set you as a reviewer or would you like someone to review the PR?

corochann:
Assigned myself. I will discuss with @mihaimorariu about this.

delta2323:
OK, thanks.

mihaimorariu:
@corochann: I would go for your second proposal. Maybe have this function inside `chainer_chemistry.utils.visualization`?

mihaimorariu:
Following the discussion with @corochann, we agreed to keep the plot script inside each example folder, for now. I will, therefore, close this PR.

PR #184, opened 2018-06-09
Remove inference_tox21.py and relevant code from the Tox21 example

delta2323:
This PR removes `inference_tox21.py` which is an old script for inference in the Tox21. Currently we have two prediction code in the Tox21 example: one that implements custom inference loop (`inference_tox21.py`), and the other using `Classifier` (`predict_tox21_with_classifier.py`). It is confusing for users that there are two examples that essentially do same things. Using `Classifier` is a recommended way for prediction, so we keep the latter (and newer) one. 

This PR also removes code that is no longer used if we remove `inference_tox21.py`


corochann:
lgtm

PR #188, opened 2018-06-12
Update test example

corochann:
Please merge after https://github.com/pfnet-research/chainer-chemistry/pull/171

Each test example script is separated in this PR.
Also, cpu/gpu can be configured as a argument of shell script.


examples/tox21/test_tox21.sh
```diff
@@ -2,19 +2,14 @@
...
+# gpu id given from first argument, default value is -1
+gpu=${1:--1}
 
-declare method_list=(nfp ggnn schnet weavenet rsgcn)
 # Preprocessor parse result must contain both pos/neg samples
 tox21_num_data=100
 
-#for method in nfp ggnn schnet weavenet rsgcn
-for (( i=0; i<${#method_list[@]}; i++ ));
+for method in nfp ggnn schnet weavenet rsgcn
```
delta2323: Do you have any specific reason why you do not use the combination of `declare` and `for` (both are OK, I'm just curious).

examples/test_examples.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -e
+
+gpu=${1:--1}
+echo Using gpu ${gpu}
+
+# Tox21
+echo --- Testing Tox21 ---
+cd tox21 && ./test_tox21.sh ${gpu} && cd ..
```
delta2323: Personally, I like `bash -x test_tox21.sh` because when the test fails, we can identify where is the problem. How do you think?

corochann: sure

examples/tox21/test_tox21.sh
corochann: When I need to use `i`, I needed to `declare`.
But now `i` is not used and I thought it is more easy to understand bash script using only `for ... in ...` notation.

corochann:
updated

examples/tox21/test_tox21.sh
delta2323: OK understood.

mottodora:
I confirmed test script runs without error.

Issue #173, opened 2018-05-17
Model evaluation with the QM9 dataset.


delta2323:
Similar to #147.

mihaimorariu:
This issue has been addressed in #189.

PR #189, opened 2018-06-18
Add evaluation script for QM9

mihaimorariu:
This PR implements #173.


examples/qm9/evaluate_models_qm9.sh
```diff
@@ -0,0 +1,15 @@
+set -eu
+
+gpu=-1
+methods=(nfp ggnn schnet weavenet rsgcn)
+methods=(nfp)
```
corochann: remove this line

examples/qm9/predict_qm9.py
```diff
@@ -155,10 +159,13 @@ def postprocess_fn(x):
     # --- evaluate ---
     # To calc loss/accuracy, we can use `Evaluator`, `ROCAUCEvaluator`
     print('Evaluating...')
-    val_iterator = SerialIterator(val, 16, repeat=False, shuffle=False)
+    test_iterator = SerialIterator(test, 16, repeat=False, shuffle=False)
     eval_result = Evaluator(
-        val_iterator, regressor, converter=concat_mols, device=args.gpu)()
+        test_iterator, regressor, converter=concat_mols, device=args.gpu,
+        name='test', eval_func=regressor.predictor)()
```
corochann: `name` is not in Evaluator.
https://github.com/chainer/chainer/blob/master/chainer/training/extensions/evaluator.py#L79

examples/qm9/train_qm9.py
```diff
@@ -71,14 +69,33 @@ def __call__(self, x0, x1):
...
             x1 = cuda.to_cpu(x1.data)
         if self.scale == 'standardize':
-            scaled_x0 = self.ss.inverse_transform(cuda.to_cpu(x0))
-            scaled_x1 = self.ss.inverse_transform(cuda.to_cpu(x1))
+            scaled_x0 = self.ss.inverse_transform(x0)
+            scaled_x1 = self.ss.inverse_transform(x1)
             diff = scaled_x0 - scaled_x1
         elif self.scale == 'none':
-            diff = cuda.to_cpu(x0) - cuda.to_cpu(x1)
+            diff = x0 - x1
```
corochann: It was originally `cuda.to_cpu(x0) - cuda.to_cpu(x1)`, to handle cupy array.
If you remove `cuda.to_cpu`, does it work on GPU??

corochann: I think above line should be

```
if isinstance(x0, Variable):
    x0 = x0.data
if isinstance(x1, Variable):
    x1 = x1.data
```
to convert Variable to array (numpy array or cupy array)
and then we should call `cuda.to_cpu()` outside of this if block.

corochann:
I could not run the code in my env due to Evaluator error, could you check again?

PR #190, opened 2018-06-18
mean_squared_error with `ignore_nan` option

mottodora:
- [x] fix test to chainer-chemistry style
- [x] add double backward tests
- [x] add documents

### DIFF
--- a/chainer_chemistry/functions/mean_squared_error.py
+++ b/chainer_chemistry/functions/mean_squared_error.py
@@ -67,7 +67,8 @@ def mean_squared_error(x0, x1, ignore_nan=False):
         x1 (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`): Input variable.
         ignore_nan (bool): If `True`, this function compute mean squared error
-            ignoring NaNs.
+            ignoring NaNs. The arithmetic mean is the sum of the non-NaN
+            elements along the axis divided by the number of whole elements.
 
     Returns:
         ~chainer.Variable:
--- a/chainer_chemistry/functions/mean_squared_error.py
+++ b/chainer_chemistry/functions/mean_squared_error.py
@@ -58,11 +58,15 @@ def mean_squared_error(x0, x1, ignore_nan=False):
     """Mean squared error function.
     This function computes mean squared error between two variables. The mean
     is taken over the minibatch. Note that the error is not scaled by 1/2.
+
     Args:
         x0 (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`): Input variable.
         x1 (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`): Input variable.
+        ignore_nan (bool): If `True`, this function compute mean squared error
+            ignoring NaNs.
+
     Returns:
         ~chainer.Variable:
             A variable holding an array representing the mean squared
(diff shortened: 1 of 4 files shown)


chainer_chemistry/functions/mean_squared_error.py
```diff
@@ -0,0 +1,75 @@
...
+import chainer.functions
+from chainer.utils import type_check
+
+
+class MeanSquaredError(function_node.FunctionNode):
+
+    """Mean squared error (a.k.a. Euclidean loss) function."""
+
+    def __init__(self, ignore_nan=False):
+        self.task_weight = None
```
corochann: Can you write 
`#TODO (mottodora): implement task weight calculation`
so that others can understand it is simply not implemented yet?

chainer_chemistry/functions/mean_squared_error.py
```diff
@@ -0,0 +1,75 @@
...
+        return diff.dot(diff) / diff.dtype.type(diff.size),
+
+    def backward(self, indexes, gy):
+        x0, x1 = self.get_retained_inputs()
+        xp = cuda.get_array_module(x0)
+        ret = []
+        diff = x0 - x1
+        if self.ignore_nan:
+            diff = chainer.functions.where(xp.isnan(diff.array),
+                                           xp.zeros_like(diff.array), diff)
```
corochann: I'm not sure but `xp.zeros_like(diff.array)` can be just 0, in that case we can skip creating (allocating memory of) maybe large zero array.

tests/functions_tests/test_mean_squared_error.py
```diff
@@ -0,0 +1,148 @@
...
+
+    loss_expect = 0.
+    x0_data = cuda.to_cpu(x0_data)
+    x2_data = cuda.to_cpu(x2_data)
+    nan_mask = numpy.invert(numpy.isnan(x2_data)).astype(x2_data.dtype)
+    for i in numpy.ndindex(x0_data.shape):
+        loss_expect += ((x0_data[i] - numpy.nan_to_num(x2_data[i])) ** 2
+                        * nan_mask[i])
+    loss_expect /= x0_data.size
+    assert numpy.allclose(loss_value, loss_expect)
```
delta2323: Why did you use `numpy.allclose` in this function, while `pytest.approx` in `check_forward`?

chainer_chemistry/functions/mean_squared_error.py
```diff
@@ -0,0 +1,77 @@
...
+from chainer import function_node
+import chainer.functions
+from chainer.utils import type_check
+
+
+class MeanSquaredError(function_node.FunctionNode):
+
+    """Mean squared error (a.k.a. Euclidean loss) function."""
+
+    def __init__(self, ignore_nan=False):
```
delta2323: IMO, what "ignore" means can differ from people to people. So, I would suggest more descriptive option name "e.g. convert_nan_to_zero" or write what this option does in detail in the document.

tests/functions_tests/test_mean_squared_error.py
mottodora: Fix this. Thank you.

tests/functions_tests/test_mean_squared_error.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def check_backward_ignore_nan(inputs):
+    x0_data, _, x2_data = inputs
+
+    def func(x0, x1):
+        return chainer_chemistry.functions.mean_squared_error(x0, x1,
+                                                              ignore_nan=True)
+    gradient_check.check_backward(func, (x0_data, x2_data), None, eps=1e-2)
+
```
delta2323: Could you add tests for backward and double backward with `ignore_nan=True` using `x0` and `x1`?

mottodora: fix

PR #196, opened 2018-06-22
Random Splitter

mottodora:
- [x] RandomSplitter
- [x] tests
- [x] documents


chainer_chemistry/dataset/splitters/random_splitter.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class RandomSplitter(BaseSplitter):
+    def split(self, dataset, seed=None, frac_train=.8, frac_valid=.1,
+              frac_test=.1):
+        numpy.testing.assert_almost_equal(frac_train + frac_valid + frac_test,
+                                          1.)
+        if seed is not None:
+            numpy.random.seed(seed)
+        perm = numpy.random.permutation(len(dataset))
```
corochann: Setting numpy random seed may have side effect to other places.
How about using `numpy.random.RandomState(seed).permutation(len(dataset))`?

corochann:
So far looks good to me for the design, I think it is ok to separate PR, and merge it in current status 
(only BaseSplitter & RandomSplitter for this PR).

corochann:
I also want to consider the case when we want to split only `train` and `val`, without test dataset.
How can we handle this case?

chainer_chemistry/dataset/splitters/base_splitter.py
```diff
@@ -1,14 +1,44 @@
...
-                                                       frac_test=frac_test)
+        train_inds, valid_inds, test_inds = self._split(dataset,
+                                                        frac_train=frac_train,
+                                                        frac_valid=frac_valid,
+                                                        frac_test=frac_test)
         if return_index:
             return train_inds, valid_inds, test_inds
         else:
-            return dataset[train_inds], dataset[valid_inds], dataset[test_inds]
+            if type(dataset) == NumpyTupleDataset:
```
corochann: use `isinstance(dataset, NumpyTupleDataset)`.

I think the different way is to introduce `converter` function as an argument, so that user can explicitly specify how to split based on `dataset` with given `indices`.

Default behavior would be...

```
def converter(dataset, indices):
    return dataset[indices]
```

and for NumpyTupleDataset, user can explicitly specify...
```
def converter_numpy_tuple_dataset(dataset, indices):
    return NumpyTupleDataset(*dataset.features[indices])
```

chainer_chemistry/dataset/splitters/base_splitter.py
```diff
@@ -1,14 +1,44 @@
...
+                train = NumpyTupleDataset(*dataset[train_inds])
+                valid = NumpyTupleDataset(*dataset[valid_inds])
+                test = NumpyTupleDataset(*dataset[test_inds])
```
corochann: use features, `NumpyTupleDataset(*dataset.features[indices])` ?

corochann:
Can you write docstring as in other chainer chemistry code (Google format)?

https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/dataset/converters.py#L5
https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/models/ggnn.py#L109

other than that, LGTM

PR #198, opened 2018-06-22
Retrieve version number in setup.py without imp

delta2323:
This PR changes to retrieve the version number of Chainer Chemistry in setup.py without using imp. Essentially same change as https://github.com/chainer/chainer/pull/4846

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
 from distutils.core import setup
-import imp
 import os
 
 from setuptools import find_packages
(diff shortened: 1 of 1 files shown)


delta2323:
@mottodora pointed out that Python has `importlib`, which is a successor of `imp`. So, I want to consider we should use it or not.

mottodora:
After offline discussion, we conclude that we shouldn't use `importlib`. Because `importlib` import library so it needs dependency.

PR #208, opened 2018-06-30
fix seed 0 in the test of models and loss functions

corochann:


corochann:
To reduce probabilistic test failure, I fixed seed for data generation.
However the test is still probabilistic (e.g., model parameter initialization is random, gpu calculation order is random), so I hope this will reduce the probabilistic failure, but it may still cause probabilistic failure.

PR #209, opened 2018-06-30
Integrate get_molnet_dataset with Splitter c6a9cc7

mottodora:
Please merge after #201 #202

PR #216, opened 2018-07-06
 Add `return_raw_df` option in get_molnet_dataset

mottodora:
Please merge after #209


corochann:
I feel adding another function is better rather than adding as an argument in `get_molnet_dataset`.
Because...
1. `preprocessor`, `frac_train` etc is not necessary to get df, so function argument will be different.
2. The behavior of `return_smiles` and `return_raw_df` is different.
When `return_smiles=True` it will ADD "smiles" information in the result, but `return_raw_df=True` will CHANGE the return value completely different from dataset.

What do you think?

mottodora:
I totally agree. I will fix. Thank you.

mottodora:
Fix and change the title of PR.

chainer_chemistry/datasets/molnet/molnet.py
```diff
@@ -129,6 +130,37 @@ def postprocess_label(label_list):
...
+    if dataset_config['dataset_type'] == 'one_file_csv':
+        df = pandas.read_csv(get_molnet_filepath(dataset_name))
+        return df
+    elif dataset_config['dataset_type'] == 'separate_csv':
+        train_df = pandas.read_csv(get_molnet_filepath(dataset_name, 'train'))
+        valid_df = pandas.read_csv(get_molnet_filepath(dataset_name, 'valid'))
+        test_df = pandas.read_csv(get_molnet_filepath(dataset_name, 'test'))
+        return train_df, valid_df, test_df
+    else:
+        raise NotImplementedError
```
corochann: Can you show message ?
'dataset_type={} is not supported'.format(dataset_config['dataset_type'])

chainer_chemistry/datasets/molnet/molnet.py
```diff
@@ -129,6 +130,37 @@ def postprocess_label(label_list):
     return result
 
 
+def get_molnet_raw_dataframe(dataset_name):
```
corochann: I think name "raw" is not necessary.

chainer_chemistry/datasets/molnet/molnet.py
```diff
@@ -129,6 +130,37 @@ def postprocess_label(label_list):
...
+    """Downloads and caches MoleculeNet dataset without preprocessing.
```
corochann: Can you update this doc?

corochann:
lgtm after test pass

PR #217, opened 2018-07-09
Relational Graph attention networks

mottodora:
Implement Graph Attention Networks
https://arxiv.org/abs/1710.10903
- [x] basic implementation
- [x] Dropout option
- [x] concat heads option
- [x] consider edge type
- [x] add unit tests
- [x] documents
- [x] example
- [x] Preprocessor


mottodora:
Actually the paper is only about update function, so I use GGNN's readout function.
We need to discuss it.

chainer_chemistry/models/gat.py
```diff
@@ -0,0 +1,169 @@
...
+        # z = functions.reshape(z, (self.heads, mb, atom, atom))
+
+        cond = adj.array.astype(xp.bool)
+        cond = xp.broadcast_to(cond, e.array.shape)
+        # TODO(mottodora): find better way to ignore non connected
+        e = functions.where(cond, e,
+                            xp.broadcast_to(xp.array(-10000), e.array.shape)
+                            .astype(xp.float32))
+        # (heads, minibatch, atom, atom)
+        alpha = functions.softmax(e)
```
corochann: softmax axis is 1, do you want to apply along minibatch??

chainer_chemistry/models/gat.py
```diff
@@ -0,0 +1,169 @@
...
+        # (minibatch * heads, atom * atom, out_dim * 2)
+        e = functions.reshape(e, (mb * self.heads, atom * atom,
+                                  self.hidden_dim * 2))
+        # (minibatch * heads, atom * atom, 1)
+        e = self.attenstion_layers[step](e)
+
+        # (minibatch, heads, atom, atom)
+        e = functions.reshape(e, (mb, self.heads, atom, atom))
+        # (heads, minibatch, atom, atom)
+        e = functions.transpose(e, (1, 0, 2, 3))
```
corochann: why `heads` axis need to be first axis?

chainer_chemistry/models/gat.py
```diff
@@ -0,0 +1,169 @@
...
+        # (heads, minibatch, atom, atom)
+        alpha = functions.softmax(e)
+        # (minibatch, heads, atom, atom)
+        alpha = functions.transpose(alpha, (1, 0, 2, 3))
+        # (minibatch, heads, atom, out_dim)
+        h = functions.transpose(h, (0, 2, 1, 3))
+        # (minibatch, heads, atom, out_dim)
+        h_new = functions.matmul(alpha, h)
+        # (minibatch, atom, out_dim)
+        h_new = functions.mean(h_new, axis=1)
```
corochann: just memo: original paper suggests both `mean` and `concat`.

chainer_chemistry/models/gat.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # concat all pairs of atom
+        # (minibatch, 1, atom, heads, out_dim)
+        h_i = functions.expand_dims(h, axis=1)
+        # (minibatch, atom, atom, heads, out_dim)
+        h_i = functions.broadcast_to(h_i, (mb, atom, atom,
+                                           self.heads, self.hidden_dim))
+        device_id = cuda.get_device_from_array(h_i.array).id
+        h_j = functions.copy(h_i, device_id)
+        h_j = functions.transpose(h_j, (0, 2, 1, 3, 4))
```
corochann: this is completely matter of preference, but I feel it is easy to understand to create `h_j` same way with `h_i` using `expand_dims` and `broadcast_to`.

chainer_chemistry/models/gat.py
mottodora: for L109
(cond's shape is `mb, atom, atom` )

corochann: Then another way is to adding axis in cond like `cond = xp.broadcast_to(cond[:, None, :, :], e.array.shape)` (using `reshape` or `expand_dims` also fine). so that we can skip using `transpose` and `transpose` back at L117 for this array `e`

mottodora: Certainly! I will fix.

chainer_chemistry/models/gat.py
mottodora: This is a mistake. Thank you!

chainer_chemistry/models/gat.py
mottodora: Sorry. I would like to make sure.
Are these codes enough?
`h_j = functions.expand_dims(h, axis=2)` 
`h_j = functions.broadcast_to(h_i, (mb, atom, atom, self.heads, self.hidden_dim))`

chainer_chemistry/models/gat.py
```diff
@@ -0,0 +1,208 @@
...
+        if self.dropout_ratio >= 0:
+            alpha = functions.dropout(alpha, ratio=self.dropout_ratio)
+
+        # before: (minibatch, atom, EDGE_TYPE, heads, out_dim)
+        # after: (minibatch, EDGE_TYPE, heads, atom, out_dim)
+        h = functions.transpose(h, (0, 2, 3, 1, 4))
+        # (minibatch, EDGE_TYPE, heads, atom, out_dim)
+        h_new = functions.matmul(alpha, h)
+        # (minibatch, heads, atom, out_dim)
+        h_new = functions.sum(h_new, axis=1)
```
corochann: let me discuss how to handle edge_type agnostic attention.

chainer_chemistry/models/gat.py
```diff
@@ -0,0 +1,208 @@
...
+        super(GAT, self).__init__()
+        n_readout_layer = n_layers if concat_hidden else 1
+        n_message_layer = n_layers
+        with self.init_scope():
+            # Update
+            self.embed = EmbedAtomID(out_size=hidden_dim, in_size=n_atom_types)
+            self.message_layers = chainer.ChainList(
+                *[GraphLinear(hidden_dim * n_heads,
+                              n_edge_types * hidden_dim * n_heads)
+                  if i > 0 and concat_heads else
```
corochann: Just a matter of preference, but how about setting
`in_channels` variable before hand with  `if i > 0 and concat_heads else` condition?

mottodora:
Can you review? @corochann

PR #220, opened 2018-07-10
add return_is_successful & modify total count behavior

corochann:
This PR 
- add `return_is_successful` option 
- modify `total_count`, `fail_count` behavior

for `SDFFileParser`. Let me discuss the specification before merge.

### DIFF
--- a/chainer_chemistry/dataset/parsers/sdf_file_parser.py
+++ b/chainer_chemistry/dataset/parsers/sdf_file_parser.py
@@ -29,7 +29,8 @@ class SDFFileParser(BaseFileParser):
         self.postprocess_fn = postprocess_fn
         self.logger = logger or getLogger(__name__)
 
-    def parse(self, filepath, return_smiles=False, target_index=None):
+    def parse(self, filepath, return_smiles=False, target_index=None,
+              return_is_successful=False):
         """parse sdf file using `preprocessor`
 
         Note that label is extracted from preprocessor's method.
@@ -42,6 +43,10 @@ class SDFFileParser(BaseFileParser):
                 `None`.
             target_index (list or None): target index list to partially extract
                 dataset. If None (default), all examples are parsed.
+            return_is_successful (bool): If set to `True`, boolean list is
+                returned in the key 'is_successful'. It represents
+                preprocessing has succeeded or not for each SMILES.
+                If set to False, `None` is returned in the key 'is_success'.
 
         Returns (dict): dictionary that contains Dataset, 1-d numpy array with
             dtype=object(string) which is a vector of smiles for each example
@@ -51,6 +56,7 @@ class SDFFileParser(BaseFileParser):
         logger = self.logger
         pp = self.preprocessor
         smiles_list = []
+        is_successful_list = []
 
         if isinstance(pp, MolPreprocessor):
             mol_supplier = Chem.SDMolSupplier(filepath)
@@ -68,7 +74,9 @@ class SDFFileParser(BaseFileParser):
                 mol = mol_supplier[int(index)]
 
                 if mol is None:
-                    total_count -= 1
+                    fail_count += 1
+                    if return_is_successful:
+                        is_successful_list.append(False)
                     continue
                 try:
                     # Labels need to be extracted from `mol` before standardize
@@ -106,10 +114,15 @@ class SDFFileParser(BaseFileParser):
                     # This is expected error that extracting feature failed,
                     # skip this molecule.
                     fail_count += 1
+                    if return_is_successful:
+                        is_successful_list.append(False)
                     continue
                 except Exception as e:
                     logger.warning('parse() error, type: {}, {}'
                                    .format(type(e).__name__, e.args))
+                    fail_count += 1
+                    if return_is_successful:
+                        is_successful_list.append(False)
                     continue
 
                 if isinstance(input_features, tuple):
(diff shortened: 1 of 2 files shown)


chainer_chemistry/dataset/parsers/sdf_file_parser.py
```diff
@@ -68,7 +74,9 @@ def parse(self, filepath, return_smiles=False, target_index=None):
                 mol = mol_supplier[int(index)]
 
                 if mol is None:
-                    total_count -= 1
+                    fail_count += 1
```
corochann: How to consider the case of "mol is None".
Previously, I was dealing this case is "not included in the counting itself".
However, now I feel it is better to consider this case as simply "fail, with counting".

corochann: Anyway, this will only affect to logging in L153. It does not affect to program input/output behavior.

tests/dataset_tests/parsers_tests/test_sdf_file_parser.py
```diff
@@ -27,6 +27,19 @@ def sdf_file(tmpdir, mols):
...
 
+@pytest.fixture()
+def sdf_file_long(tmpdir):
+    """SDFFile with long smiles (ccc...)"""
+    fname = os.path.join(str(tmpdir), 'test_long.sdf')
+    writer = Chem.SDWriter(fname)
+    for smiles in ['CCCCCCCCCCCC', 'CN=C=O', 'CCCCCCCCCCCCCCCC',
+                   'Cc1ccccc1', 'CC1=CC2CC(CC1)O2']:
+        mol = Chem.MolFromSmiles(smiles)
+        writer.write(mol)
```
corochann: I dont know how to create "invalid" sdf file.
So I made sdf file with larger size of smiles (more than 10 atoms).

tests/dataset_tests/parsers_tests/test_sdf_file_parser.py
```diff
@@ -96,15 +115,34 @@ def test_sdf_file_parser_target_index(sdf_file, mols):
     assert smiles[1] == 'CC1=CC2CC(CC1)O2'
 
 
+def test_sdf_file_parser_return_is_successful(sdf_file_long, mols):
+    """test `labels` option and retain_smiles=True."""
+    preprocessor = NFPPreprocessor(max_atoms=10)
```
corochann: specifying max_atoms=10 to see if `is_successful` work correctly or not.
I know that this depends on the `preprocessor` behavior, which is not good way of unittesting...

chainer_chemistry/dataset/parsers/sdf_file_parser.py
mottodora: LGTM.
BTW, why did you choice "not included in the counting itself"?
Is there any specific reason to do it?

tests/dataset_tests/parsers_tests/test_sdf_file_parser.py
mottodora: If you don't want to make invalid 'sdf', this code can make it.(1st mol is broken.)

```
with open('test_long.sdf', 'r') as f:
    lines = f.readlines()
del(lines[10])
with open('test_long.sdf', 'w') as f:
    f.writelines(lines)
```

tests/dataset_tests/parsers_tests/test_sdf_file_parser.py
```diff
@@ -27,6 +27,19 @@ def sdf_file(tmpdir, mols):
...
+@pytest.fixture()
+def sdf_file_long(tmpdir):
+    """SDFFile with long smiles (ccc...)"""
+    fname = os.path.join(str(tmpdir), 'test_long.sdf')
+    writer = Chem.SDWriter(fname)
+    for smiles in ['CCCCCCCCCCCC', 'CN=C=O', 'CCCCCCCCCCCCCCCC',
+                   'Cc1ccccc1', 'CC1=CC2CC(CC1)O2']:
+        mol = Chem.MolFromSmiles(smiles)
+        writer.write(mol)
+    print('fname', fname)
```
mottodora: debug print?

chainer_chemistry/dataset/parsers/sdf_file_parser.py
corochann: Because when mol is None, this is not even included in preprocessing trial, I thought at the first time it is better to just exclude it.
But now if we want to refer `is_successful` array, i think containing all information is better.

tests/dataset_tests/parsers_tests/test_sdf_file_parser.py
corochann: I see thanks. but removing middle of the line (1st mol within total 3 mols etc.)  is difficult, so may i proceed for current PR?

tests/dataset_tests/parsers_tests/test_sdf_file_parser.py
corochann: fixed

chainer_chemistry/dataset/parsers/sdf_file_parser.py
mottodora: I see. Thank you.

tests/dataset_tests/parsers_tests/test_sdf_file_parser.py
mottodora: OK.

PR #223, opened 2018-07-12
Inputsize invariant rsgcn

corochann:
This is a test to explicitly check that model forward result is same even if input size is bigger (with atom & adjacency matrix filled with 0).

### DIFF
--- a/chainer_chemistry/utils/extend.py
+++ b/chainer_chemistry/utils/extend.py
@@ -1,4 +1,5 @@
 from collections import Iterable
+from logging import getLogger
 
 import numpy
 import six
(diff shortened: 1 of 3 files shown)


chainer_chemistry/utils/extend.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def extend_adj(adj, out_size, axis=None, value=0):
+    axis = axis or [-1, -2]
+    return extend_arrays_to_size(
+        adj, out_size=out_size, axis=axis, value=value)
+
+
+def extend_arrays_to_size(arrays, out_size, axis=-1, value=0):
+    """Extend size of `node` array
```
mottodora: Do you think users will use `extend_arrays_to_size` function? Anyway, this docstring  is for `extend_node`.

chainer_chemistry/utils/extend.py
```diff
@@ -0,0 +1,66 @@
...
+        a = [a]
+    return a
+
+
+def extend_node(node, out_size, axis=-1, value=0):
+    return extend_arrays_to_size(
+        node, out_size=out_size, axis=axis, value=value)
+
+
+def extend_adj(adj, out_size, axis=None, value=0):
```
mottodora: Please add docstring.

chainer_chemistry/utils/extend.py
```diff
@@ -0,0 +1,66 @@
...
+
+def _to_list(a):
+    if isinstance(a, Iterable):
+        a = list(a)
+    else:
+        a = [a]
+    return a
+
+
+def extend_node(node, out_size, axis=-1, value=0):
```
mottodora: Please add docstring.

chainer_chemistry/utils/extend.py
```diff
@@ -0,0 +1,66 @@
...
+    Returns (numpy.ndarray): extended `node` array, extended place is filled
+        with `value`
+
+    """
+    batch_size = len(arrays)
+
+    in_shape = _to_list(arrays[0].shape)
+    out_shape = [batch_size] + in_shape
+
+    axis = _to_list(axis)
```
mottodora: I feel if axis contains `0`, raising warning is kind  to user. But it is not necessary to solve.

chainer_chemistry/utils/extend.py
corochann: maybe yes, I update docstring.

chainer_chemistry/utils/extend.py
corochann: ok

PR #227, opened 2018-07-13
Change RDKit version in Travis CI Test

mottodora:
Many many errors are happened recently in Travis CI. This PR explore why it is happened.


mottodora:
I confirm these errors are caused by RDKit latest version.

PR #228, opened 2018-07-13
[WIP] no_compat: update NFP model to be input size invariant

corochann:
Implementation update & test done.
But since it is no-compat, I would like to discuss API before merge.


corochann:
passed to https://github.com/pfnet-research/chainer-chemistry/pull/296, let me close it.

PR #229, opened 2018-07-13
[WIP] input size invariant implementation for GGNN

corochann:
input size invariant implementation for GGNN
We should discuss API, same with https://github.com/pfnet-research/chainer-chemistry/pull/228


corochann:
passed to https://github.com/pfnet-research/chainer-chemistry/pull/297, let me close it.

PR #233, opened 2018-07-15
Using recommended metrics in train_molnet.py

mottodora:


mottodora:
Now there are some problems in this PR.
1. we can't save models because this implementation use defined function in 
`AttributeError: Can't pickle local object 'main.<locals>.loss_fun'`
https://github.com/mottodora/chainer-chemistry/blob/f35b10ad6e8bd19dba0cc1eb1d1338f51fefcdcf/examples/molnet/train_molnet.py#L179-L180

2. In some datasets, it seems Evaluator raises ValueError Because

corochann:
1. as written in the error message
> AttributeError: Can't pickle local object 'main.<locals>.loss_fun'

you can try defining function in *global scope* and try pickle again?

2. 
> ('Only one class present in y_true. ROC AUC score is not defined in that case.',)
It means that dataset contains *only one of positive or negative label*. This is caused by may be too imbalanced dataset, but is it really expected behavior??

corochann:
could you resolve conflict?

mottodora:
Update.

examples/molnet/train_molnet.py
```diff
@@ -189,20 +203,24 @@ def get_dataset_paths(cache_dir, num_data):
...
-        # print_report_targets.append('train/main/roc_auc')
-        trainer.extend(ROCAUCEvaluator(
-            val_iter, model, device=args.gpu, eval_func=predictor,
-            converter=concat_mols, name='val', raise_value_error=False))
-        print_report_targets.append('val/main/roc_auc')
+    for metric_name, metric_fun in metrics.items():
+        if isinstance(metric_fun, types.FunctionType):
+            print_report_targets.append('main/' + metric_name)
+            print_report_targets.append('validation/main/' + metric_name)
+        elif BatchEvaluator in inspect.getmro(metric_fun):
```
corochann: what is the difference from `isinstance(metric_fun, BatchEvaluator)`?

examples/molnet/train_molnet.py
```diff
@@ -189,20 +203,24 @@ def get_dataset_paths(cache_dir, num_data):
...
+            #     converter=concat_mols, name='train',
+            #     raise_value_error=False))
+            # print_report_targets.append('train/main/roc_auc')
+            trainer.extend(metric_fun(
+                val_iter, model, device=args.gpu, eval_func=predictor,
+                converter=concat_mols, name='val',
+                raise_value_error=False))
+            print_report_targets.append('val/main/' + metric_name)
+        else:
+            raise TypeError
```
corochann: Could you add some message?

PR #234, opened 2018-07-15
Loosen stratified split test.

mottodora:
I set strict value. But it fails too much.


corochann:
I think the other way is fix seed for deterministic behavior.

PR #235, opened 2018-07-19
Fix config in clearance dataset

mottodora:


corochann:
We have noticed the dataset file itself has changed recently.
http://deepchem.io.s3-website-us-west-1.amazonaws.com/datasets/clearance.csv

PR #249, opened 2018-08-03
Add example for custom dataset

mihaimorariu:
This PR implements the example scripts for using a custom dataset.

PR #251, opened 2018-08-05
Prediction by pretrained model in own_dataset

kazuyaujihara:
Usage: Train like `python train.py [options]` and predict like `train.py [options] --predict c1ccccc1`.

The file name, train.py, is not proper after this patch.


corochann:
Thank you for creating PR, but we had the same PR before, so please let me close this PR.
https://github.com/pfnet-research/chainer-chemistry/pull/249

You are welcome for the other PRs in the future!

PR #260, opened 2018-10-03
moved construct_discrete_edge_matrix to common

corochann:
Because this function may be used in other preprocessor.


corochann:
Added no-compat because the method's namespace will be changed.

PR #264, opened 2018-10-04
always use canonical smiles as returned smiles

corochann:


corochann:
Let me discuss this design is better or not before merge it.

PR #266, opened 2018-10-15
Fix test_example.sh

mottodora:


corochann:
`test_own_dataset.sh` is deleted in this PR
https://github.com/pfnet-research/chainer-chemistry/pull/249/files#diff-3bc65f8c42232e3f1f4fb3228b7ef760

I think the purpose of `test_own_dataset.sh` and `evaluate_own_dataset.sh` is different, so we should add `test_own_dataset.sh` again.

PR #268, opened 2018-10-19
remove assert check for canonical smiles

corochann:


corochann:
Actually this assert fails when mol is kekulized, but it is acceptable behavior that canonical smiles and Chem.MolToSmiles(mol) is different for kekulized mol

PR #271, opened 2018-11-16
Force the regressor to report numpy arrays

mihaimorariu:
When training a network on the CPU and using the `Evaluator` class to output the `MAE` or `MSE`, the reported values are scalars. However, when running the same code on the GPU, the reported values become arrays (and the whole output cannot be serialized into a JSON file). There seems to be a bug in Chainer's reporter class, since the conversion to a scalar is not automatically done, in this case. I will report this to the Chainer team, but this PR should be a temporary fix until the issue is settled.


corochann:
Thank you for PR, Can you check travis?

corochann:
Can you check travis? Maybe `from chainer import cuda` is not imported @mihaimorariu

mihaimorariu:
Thank you! I will fix it, I think I know what the issue is.

mihaimorariu:
All checks have passed, can you please review now?

PR #276, opened 2018-11-27
[WIP] Zinc 250k dataset support

corochann:
Implementation almost done, but I'm final checking the function behavior.


tests/datasets_tests/test_zinc.py
```diff
@@ -0,0 +1,113 @@
...
+    assert atoms.dtype == numpy.int32
+    assert label.ndim == 1
+    assert label.shape[0] == ZINC250K_NUM_LABEL
+    assert label.dtype == numpy.float32
+
+    # --- Test number of dataset ---
+    assert len(dataset) == ZINC250K_NUM_DATASET
+
+
+def test_get_zinc_smiles():
```
mottodora: Please add `@pytest.mark.slow`

corochann: This is intentional. I used `target_index` so this test is not slow.

corochann:
Thanks!

Issue #281, opened 2018-12-07
Report metrics as scalars


mihaimorariu:
When training a network on the CPU and using the `Evaluator` class to output the `MAE` or `MSE`, the reported values are scalars. However, when running the same code on the GPU, the reported values become arrays (and the whole output cannot be serialized into a `JSON` file). There seems to be a bug in Chainer's reporter class, since the conversion to a scalar is not automatically done, in this case. I will report this to the Chainer team, but this PR should be a temporary fix until the issue is settled.

PR #282, opened 2018-12-07
Report metrics as scalars

mihaimorariu:
This PR addresses #281.


chainer_chemistry/models/prediction/regressor.py
```diff
@@ -112,13 +123,19 @@ def __call__(self, *args, **kwargs):
...
         self.y = self.predictor(*args, **kwargs)
         self.loss = self.lossfun(self.y, t)
-        reporter.report({'loss': self.loss}, self)
+
+        # When the reported data is a numpy array, the loss and metrics values
+        # are scalars. When the reported data is a cupy array, sometimes the
+        # same values become arrays instead. This seems to be a bug inside the
+        # reporter class, which needs to be addressed and fixed. Until then,
+        # the reported values will be converted to numpy arrays.
+        reporter.report({'loss': Variable(cuda.to_cpu(self.loss.data))}, self)
```
corochann: Do we need to wrap it to Variable?

mihaimorariu: Some tests were failing when I tried to report the value itself. Maybe I did something wrong along the way. I will double check and remove `Variable` if everything works.

PR #285, opened 2018-12-12
take auto sum when output_var.size is not 1.

corochann:


corochann:
This PR is for when one want to backward output_var other than `loss`.

For example, if we want to calculate the gradient of output, not loss we need to write `eval_fun` as follows.
```
def eval_fun(x, adj, t):
    pred =  predictor(x, adj)
    pred_summed = F.sum(pred)
    return pred_summed
```

But usually, we forget to write `F.sum`, and try to return `pred` which is multi-dimensional along batch size and backward does not work.

corochann:
TODO: test

PR #291, opened 2018-12-17
bug fix: consider the case when metrics returns non-Variable case.

corochann:
resolves https://github.com/pfnet-research/chainer-chemistry/issues/289

Metrics function can return normal array instead of Variable, so `self._convert_to_scalar(value(self.y, t).data` is not ok (when the returned value is array, `.data` refer memoryview instead of actual value).


corochann:
I got that error before PR (current status), and I hope it should be fixed after merging this PR.

memoryview is numpy array's `data` attribute.

mihaimorariu:
As discussed in private, I will merge this and look into why the example scripts crash later.
Looks good to me otherwise, thank you for the PR!

PR #292, opened 2018-12-17
Fix ggnn

natsukium:
I have refactored GGNN model which is able to handle args; step, concat_hidden and weight_tying.
Due to this, GGNNUpdate and GGNNReadout modules have no state of network.

### DIFF
--- a/chainer_chemistry/links/readout/ggnn_readout.py
+++ b/chainer_chemistry/links/readout/ggnn_readout.py
@@ -11,9 +11,6 @@ class GGNNReadout(chainer.Chain):
         out_dim (int): dimension of output feature vector
         hidden_dim (int): dimension of feature vector associated to
             each atom
-        n_layers (int): number of layers
-        concat_hidden (bool): If set to True, readout is executed in
-            each layer and the result is concatenated
         nobias (bool): If ``True``, then this function does not use
             the bias
         activation (~chainer.Function or ~chainer.FunctionNode):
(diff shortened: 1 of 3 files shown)


chainer_chemistry/links/update/ggnn_update.py
```diff
@@ -13,32 +13,24 @@ class GGNNUpdate(chainer.Chain):
...
-        n_layers (int): number of layers
         n_atom_types (int): number of types of atoms
         num_edge_type (int): number of types of edge
-        weight_tying (bool): enable weight_tying or not
     """
 
-    def __init__(self, hidden_dim=16, n_layers=4,
-                 n_atom_types=MAX_ATOMIC_NUM, num_edge_type=4,
-                 weight_tying=True):
+    def __init__(self, hidden_dim=16, n_atom_types=MAX_ATOMIC_NUM,
```
mottodora: `n_atom_types` is not necessary for GGNN update.

chainer_chemistry/models/ggnn.py
```diff
@@ -65,22 +65,22 @@ def __call__(self, atom_array, adj):
         Returns:
             ~chainer.Variable: minibatch of fingerprint
         """
-        # reset state
-        self.update_layer.reset_state()
```
corochann: Can you implement `reset_state` method for GGNN, and reset all the update_layers in the `reset_state` method?

chainer_chemistry/models/ggnn.py
```diff
@@ -65,22 +65,22 @@ def __call__(self, atom_array, adj):
...
         if atom_array.dtype == self.xp.int32:
             h = self.embed(atom_array)  # (minibatch, max_num_atoms)
         else:
             h = atom_array
         h0 = functions.copy(h, cuda.get_device_from_array(h.data).id)
         g_list = []
         for step in range(self.n_layers):
-            h = self.update_layer(h, adj, step)
+            message_layer_index = 0 if self.weight_tying else step
+            self.update_layers[message_layer_index].reset_state()
```
corochann: please reset inside `reset_state` method.

chainer_chemistry/models/ggnn.py
```diff
@@ -28,23 +28,23 @@ class GGNN(chainer.Chain):
...
+            self.update_layers = chainer.ChainList(*[GGNNUpdate(
+                hidden_dim=hidden_dim, n_atom_types=self.NUM_EDGE_TYPE)
+                for _ in range(n_message_layer)])
             # Readout
-            self.readout_layer = GGNNReadout(
+            self.readout_layers = chainer.ChainList(*[GGNNReadout(
                 out_dim=out_dim, hidden_dim=hidden_dim,
-                n_layers=n_readout_layer, concat_hidden=concat_hidden,
-                activation=functions.identity)
+                activation=functions.identity) for _ in range(n_readout_layer)]
```
corochann: Could you add activation as GGNN argument, and set it?

chainer_chemistry/links/readout/ggnn_readout.py
```diff
@@ -11,43 +11,31 @@ class GGNNReadout(chainer.Chain):
...
-        index = step if self.concat_hidden else 0
         # h, h0: (minibatch, atom, ch)
         h1 = functions.concat((h, h0), axis=2) if h0 is not None else h
 
-        g1 = functions.sigmoid(self.i_layers[index](h1))
-        g2 = self.activation(self.j_layers[index](h1))
+        g1 = functions.sigmoid(self.i_layer(h1))
+        g2 = self.activation(self.j_layer(h1))
         # sum along atom's axis
         g = self.activation(functions.sum(g1 * g2, axis=1))
```
corochann: Please remove to apply activation here.

It was not existed.
https://github.com/pfnet-research/chainer-chemistry/blob/v0.4.0/chainer_chemistry/models/ggnn.py#L105

natsukium: According to the paper, https://arxiv.org/abs/1511.05493, tanh functions are applied in eq.7.
It also says that "The tanh functions can also be replaced with the identity".
I think you are able to select which functions to use.

natsukium:
I improve GGNN based on comment without activation.
Could you check it please?

chainer_chemistry/models/ggnn.py
```diff
@@ -28,26 +28,27 @@ class GGNN(chainer.Chain):
     """
     NUM_EDGE_TYPE = 4
 
-    def __init__(self, out_dim, hidden_dim=16,
-                 n_layers=4, n_atom_types=MAX_ATOMIC_NUM, concat_hidden=False,
-                 weight_tying=True):
+    def __init__(self, out_dim, hidden_dim=16, n_layers=4,
+                 n_atom_types=MAX_ATOMIC_NUM, activation=functions.identity,
+                 concat_hidden=False, weight_tying=True):
```
corochann: sorry, could you add `activation` as final argument to keep backward compatibility as much as possible?

chainer_chemistry/links/readout/ggnn_readout.py
corochann: I see... in that case I think it is nice to to separate activation function for line 38 and 40, so that one may use tanh in only one place. what do you think?

corochann:
LGTM, let me separate activation function in other PR.

PR #297, opened 2018-12-20
GGNN input size invariant: add is_real_node argument

corochann:


corochann:
take over from https://github.com/pfnet-research/chainer-chemistry/pull/229

corochann:
resolved conflict

PR #305, opened 2019-01-21
refactor general readout: move to links.

corochann:


corochann:
Could you review? @natsukium

natsukium:
LGTM but it is better to note down in docstring that GeneralReadout class has no parameter and is interpreted as `Link`.
I think users and other contributors can avoid confusion.

natsukium:
I think it's good.
Thank you!

PR #306, opened 2019-01-21
softmax with mask

corochann:


chainer_chemistry/functions/activation/softmax.py
```diff
@@ -0,0 +1,12 @@
+from chainer import functions
+
+
+def softmax(x, axis=1, mask=None, mask_value=1e10):
```
mottodora: Please add docstring.

chainer_chemistry/functions/activation/softmax.py
```diff
@@ -0,0 +1,12 @@
...
+    if mask is None:
+        h = x
+    else:
+        if x.shape != mask.shape:
```
mottodora: a bit strict? But we can fix later.

mottodora: We can check the range of value. (0. <= x <= 1.)
But we can add later.

mottodora: Sorry. I would like to suggest we can broadcast if the dimension(of input) is consistent. So please ignore this comment now.

corochann: yeah may be we can broadcast internally but i am not sure about all use case yet.

chainer_chemistry/functions/activation/softmax.py
corochann: ok.

chainer_chemistry/functions/activation/softmax.py
corochann: let us add later.

PR #318, opened 2019-02-01
Flow Scaler

YoshikawaMasashi:
I am implementing "Flow Scaler", that scale data into the normal distribution.
This Scaler uses a technique named "flow".

image


corochann:
Is the implementation follow from some paper? If so, please cite inside `docstring`.

I also wonder `BaseScaler` should have `inverse_transform` method or not. (which means, this scaler should be BaseScaler or other class maybe Normalizer class?)
The scaler without `inverse_transform` is still useful for input feature normalization, but not applicable for label scaling.

LGTM except that.

PR #334, opened 2019-02-23
[WIP] Cora dataset

natsukium:


corochann:
How do we proceed this one?
I want to merge it if it can at least download cora dataset correctly.

corochann:
I would like to take over this PR in https://github.com/pfnet-research/chainer-chemistry/pull/402.
I re-used some methods which you implemented, thank you @natsukium !!

PR #336, opened 2019-02-28
Move setup predictor to Library

natsukium:


natsukium:
I added kwargs to set_up_predictor.
Please check it.

PR #338, opened 2019-03-05
fix import & docstring

corochann:


corochann:
This is just fixing import, removing unnecessary import, and fix docstring?
Could you quickly check and merge after test pass? @mottodora

PR #339, opened 2019-03-11
Fix the type of return value from R2 Score Evaluator

mottodora:


natsukium:
Could you fix `test_r2_score_evaluator.py` too?
Thank you for fixing my mistake.

natsukium:
These indices is unnecessary.

https://github.com/pfnet-research/chainer-chemistry/blob/6de82de4cafd38a9826e68771432bff6ab01edc5/tests/training_tests/extensions_tests/test_r2_score_evaluator.py#L96

mottodora:
Fixed.

PR #342, opened 2019-03-18
Implement SparseGGNN

ir5:
This PR adds new implementation: GGNN for sparse matrix inputs.

* I slightly modified the implementation of `GGNNUpdater`. Now the updater can accept `CooMatrix` as well as dense `ndarray`.
* The original `GGNN` is unchanged.
* `SparseGGNN` is newly introduced.
* Several tests are added.

### DIFF
--- a/chainer_chemistry/links/update/ggnn_update.py
+++ b/chainer_chemistry/links/update/ggnn_update.py
@@ -4,6 +4,7 @@ from chainer import links
 
 import chainer_chemistry
 from chainer_chemistry.links.connection.graph_linear import GraphLinear
+from chainer_chemistry.utils import is_sparse
 
 
 class GGNNUpdate(chainer.Chain):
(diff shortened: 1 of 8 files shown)


corochann:
Could you fix tests?

 - Below chainer<=4, CooMatrix does not exist, so please just skip test in that case.
   Error message: `AttributeError: module 'chainer.utils' has no attribute 'CooMatrix'`

- For python 3.5, seems numpy produces error... Could you check it?

```
df = np.diff(sorted_pos, prepend=sorted_pos[0]) != 0
E       TypeError: diff() got an unexpected keyword argument 'prepend'
```

corochann:
I would like to merge it after test is fixed.
After merged, I would like to further refactor sparse model support as follows if possible:

 - make preprocessor produces `CooMatrix`
 - support converter to make batch for `CooMatrix`
 - Sparse model just receives batched `CooMatrix`, instead of separated `data`, `row`, `col`.

ir5:
I modified the code to skip tests of sparse models when required modules are not available.

It seems that CI build have failed because of `test_standard_scaler.py`. Why?

ir5:
I understood, some behavior has been changed since numpy==1.16.3 and thus `serializers.load_npz` is not working properly. https://github.com/pfnet-research/chainer-chemistry/pull/349#issuecomment-487424549

corochann:
https://github.com/pfnet-research/chainer-chemistry/pull/347/commits/f690a76e24e853e1a4177ed0336863b34be7f499

I think numpy behavior has changed since 1.16.3, and chainer serializer is not working well now.

PR #346, opened 2019-04-08
Schnet add h

mist714:
Information of hydrogen atom is necessary to calculate three-dimensional distance accurately.
Therefore, we added a code to calculate the distance by adding hydrogen once, and a code to remove it again if the number of hydrogen increases.


corochann:
please add `add_Hs` option as kwargs for the method.

Can we check `addHs` is already called or not by checking `mol`'s attrib??
If possible please call `add_Hs` only when `mol` 's addHs is not called yet.

Please remove assert check.

corochann:
Can you update code @mist714 ?

mottodora:
fix by #390

corochann:
done in #390

PR #348, opened 2019-04-28
Scaler axis

corochann:
StandardScaler with more than 3-dim array


chainer_chemistry/links/array/shape_transformer_2d.py
```diff
@@ -0,0 +1,57 @@
+import chainer
+from chainer import functions
+
+
+class ShapeTransformer2D(chainer.Link):
```
mottodora: `ShapeTransformerTo2D` is better?

corochann:
renamed to ShapeTransformerTo2D

PR #349, opened 2019-04-28
experimental commit

corochann:
DO NOT MERGE.

this is to check scaler_serialize test.


corochann:
I understand that issue comes from numpy behavior has changed from version 1.16.3.
Now we need to pass `allow_pickle=True` explicitly in order to load `None` object.

Reference: https://stackoverflow.com/questions/55824625/how-to-fix-object-arrays-cannot-be-loaded-when-allow-pickle-false-in-the-sketc

This should be fixed in chainer serializer.load_npz.

PR #352, opened 2019-05-05
support Path instance json encoding

corochann:


corochann:
updated to skip pathlib for python v2.7, please review.

Issue #367, opened 2019-07-19
Refactoring Graph Warp Module


corochann:
Need further refactoring: see comment of https://github.com/pfnet-research/chainer-chemistry/pull/340

PR #343, opened 2019-03-20
Add unified GraphConvModel

mottodora:
- [x] draft model
- [x] Change all update and readout argument
- [x] Test all Update and Readout function

PR #368, opened 2019-07-31
Unify GWM, modules args/kwargs

corochann:


corochann:
Take over from
 - https://github.com/pfnet-research/chainer-chemistry/pull/343
 - https://github.com/pfnet-research/chainer-chemistry/issues/367

Direction
 - Refactored to Implement all GWM models in `gwm` level modules.
    GWMGraphConvModel supports many types of combination, but we may not support all combination.
 - Refactored args/kwargs to use common name for all model implementations.

tests/links_tests/update_tests/test_mpnn_update.py
```diff
@@ -148,8 +153,10 @@ def test_backward_gpu(update, data):
     # type: (MPNNUpdate, Tuple[numpy.ndarray, numpy.ndarray, numpy.ndarray, numpy.ndarray]) -> None  # NOQA
     atom_data, adj_data, y_grad = map(cuda.to_gpu, data[:3])
     update.to_gpu()
+    # gradient_check.check_backward(update, (atom_data, adj_data), y_grad,
```
mottodora: Not necessary? If so, please remove after this PR.

PR #370, opened 2019-08-01
use label_scaler

corochann:
Use label_scaler.inverse_transform for both train/predict
It makes code very easy to understand.

Computational time for always postprocess `inverse_transform` is quite small and i think this is acceptable.


corochann:
no-compat because `GraphConvPredictor.__call__` behavior will change to apply inverse_transform.

PR #378, opened 2019-08-16
Add flexci

mottodora:


corochann:
LGTM for this PR.

Want to discuss following:

 - Which chainer version to test? Do we specify in dockerfile?
     I feel nice to test with v6.0.0 and latest pre-release version. (Maybe okay to test only with py36.)
 - Nice to specify rdkit version in dockerfile (current latest version 2019.03.2.0 as recommended version)
 - Test ChainerX build version and non-build version. (Maybe okay to test only with py36.)
    - currently tested with only non-build version, and maybe this is enough for now.
 - Test CPU-only version (No CUDA and cupy)?

mottodora:
* I would like to test chainer v6 latest and pre-release.
* It is good to specify RDKit version.
* I would like to make chainerx test in another PR.
* We need to test in CPU environment.

PR #379, opened 2019-08-19
add mark.slow flag in test_flow_scaler

corochann:
test_flow_scaler takes time because it internally trains model inside `fit` method.
This PR adds "slow" flag to skip test for CI.

Instead I added simple test to only check to run each method.


corochann:
please review & merge

PR #388, opened 2019-09-10
return svg text in MolVisualizer

corochann:


corochann:
No-compat: The returned value will change to raw `svg` text format.

When using on jupyter notebook, user need to write below explicitly.

```
from IPython.core.display import SVG
SVG(svg.replace('svg:', ''))
```

chainer_chemistry/saliency/visualizer/mol_visualizer.py
```diff
@@ -46,9 +46,24 @@ def __init__(self, logger=None):
 
     def visualize(self, saliency, mol, save_filepath=None,
                   visualize_ratio=1.0, color_fn=red_blue_cmap,
-                  scaler=abs_max_scaler, legend=''):
+                  scaler=abs_max_scaler, legend='', raise_import_error=True
```
dtaniwaki: I think `raise_import_error` should be `False` by default if we follow the original behavior of saving png files.

corochann: okay i will fix it to set False as default.

PR #393, opened 2019-09-12
Kekulize

mist714:
bug fix: add kekulize and addHs to preprocessor


corochann:
prepare_smiles_and_mol(mol) is called inside parser

corochann:
Let me close this PR. Problem is not in preprocessor.

PR #396, opened 2019-09-18
Crystal Support 

nd-02110114:
## DONE
- [x] implement MEGNet Model
- [x] support Material Project Dataset
- [x] refactoring converter method (each model has each converter method)
- [x] migrate CGCNN Model
- [x] confirm test
  - [x] pass `examples/test_example.sh` 
  - [x] pass `pytest -m "not (gpu or slow)" tests`

The detail description is [here](https://github.com/nd-02110114/chainer-chemistry/pull/1)

## TODO
- [ ] add test (WIP)
- [ ] fix path for Material Project Dataset

PR #403, opened 2019-10-03
MP auto download

corochann:
From PR https://github.com/pfnet-research/chainer-chemistry/pull/396 

 - [x] include changes of current master branch (dataset class are added to support converter)
 - [x] write download script of materials project dataset
 - [x] modify data_dir path etc.
 - [x] include citation/license properly
 - [x] check code runs
 - [ ] merge to master branch --> this PR


corochann:
## Implementation design for `converter`
Currently it is difficult to unify deciding which `converter` to use.

Current implementation is as follows:
 - `NumpyTupleDataset` ->use `converter_method_dict` to switch converter based on model
 - Other dataset (PaddingDataset, SparseDataset) -> use `dataset.converter` to use converter on dataset.

## Check
I confirmed following example runs.

 - `bash -x test_mp.sh 0`: train megnet & cgcnn with mp dataset, on GPU.
 - `bash -x test_qm9.sh 0`: train megnet works well, as well as sparse `gin_sparse` model.
 - `bash -x test_molnet.sh` wtih modified `--method megnet`: train megnet with molnet.
 - `bash -x test_tox21.sh 0`: training code works without backward compatibility break.

corochann:
I dropped testing py27 & py35 as well. When `pymatgen` is installed, it seems `pandas` version will change and work only when py36~ is used.

corochann:
Let me re-organize code and submit PR later.

PR #425, opened 2020-08-14
Refactor wle

corochann:


chainer_chemistry/dataset/parsers/data_frame_parser.py
```diff
@@ -162,7 +162,8 @@ def parse(self, df, return_smiles=False, target_index=None,
         else:
             raise NotImplementedError
 
-        smileses = numpy.array(smiles_list) if return_smiles else None
+        smileses = numpy.array(
+            smiles_list, dtype=object) if return_smiles else None
```
corochann, line 166: We need to specify dtype=object for new numpy version.

chainer_chemistry/datasets/molnet/molnet.py
```diff
@@ -113,12 +113,7 @@ def postprocess_label(label_list):
             raise TypeError("split must be None, str or instance of"
                             " BaseSplitter, but got {}".format(type(split)))
 
-        if isinstance(splitter, ScaffoldSplitter):
-            get_smiles = True
-        else:
-            get_smiles = return_smiles
-
-        if isinstance(splitter, DeepChemScaffoldSplitter):
+        if isinstance(splitter, (ScaffoldSplitter, DeepChemScaffoldSplitter)):
```
corochann, line 116: bugfix, when ScaffoldSplitter, get_smiles was False, but we want True.

PR #427, opened 2020-10-23
Remove typing dependency

mottodora:
We don't support python2 and it has difficulty in installing in python3.7.


corochann (approved):
LGTM

